PR #3, opened 2020-02-24
Adding AI-Library manifests

crobby:
Adding AI Library manifests to install the AI Library Operator.
To test this, you'll need to also have Seldon installed (always a dep of ai-library).
Here's the sample CR for AI-Library: https://gitlab.com/opendatahub/ai-library/-/blob/master/operator/deploy/crds/ailibrary_v1alpha1_ailibrary_cr.yaml
You can set most of the models there to "enabled: false" if you're tight on resources. When you create that CR, you should start to see the SeldonDeployments for those enabled: true models show-up.


.gitignore
```diff
@@ -0,0 +1,3 @@
+kfdef/.cache/manifests
+kfdef/.cache/odh-manifests
+kfdef/kustomize/
```
vpavlin: Missing new line

ai-library/base/kustomization.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+- crd.yaml
+- service-account.yaml
+- role.yaml
+- role-binding.yaml
+- operator.yaml
+namespace: kubeflow
```
vpavlin: I am planning on using namespace `opendatahub` for ODH components, what do you think?

kfdef/kfctl_openshift.yaml
```diff
@@ -0,0 +1,19 @@
...
+spec:
+  applications:
+  - kustomizeConfig:
+      repoRef:
+        name: odh-manifests
+        path: ai-library
+    name: ai-library    
+  repos:
+  - name: manifests
+    uri: /home/croberts/src/manifests
```
vpavlin: NOTE: we need to fix this before merge

ai-library/base/kustomization.yaml
crobby: That's probably a fine idea

kfdef/kfctl_openshift.yaml
crobby: yep

kfdef/kfctl_openshift.yaml
```diff
@@ -0,0 +1,19 @@
...
+      repoRef:
+        name: odh-manifests
+        path: ai-library
+    name: ai-library    
+  repos:
+  - name: manifests
+    uri: https://github.com/opendatahub-io/manifests/tarball/v0.7-branch-openshift
+  - name: odh-manifests
+    uri: https://github.com/opendatahub-io/odh-manifests/tarball/v0.7-branch-openshift
+  version: v0.7-branch-openshift
```
nakfour: did we decide to use the 0.7 branch or 1.0 Kubeflow?

crobby: Not sure we formalized our final answer on that.  Happy to change this if needed.

nakfour:
@crobby can you please add instructions to test this.

ai-library/base/operator.yaml
```diff
@@ -0,0 +1,49 @@
...
+    spec:
+      serviceAccountName: ailibrary-operator
+      containers:
+        - name: ansible
+          command:
+          - /usr/local/bin/ao-logs
+          - /tmp/ansible-operator/runner
+          - stdout
+          image: "quay.io/croberts/ai-library-operator"
+          imagePullPolicy: "Always"
```
nakfour: Can we place this image in the opendatahub quay account

ai-library/base/operator.yaml
```diff
@@ -0,0 +1,49 @@
...
+          volumeMounts:
+          - mountPath: /tmp/ansible-operator/runner
+            name: runner
+            readOnly: true
+        - name: ailibrary-operator
+          image: "quay.io/croberts/ai-library-operator"
+          imagePullPolicy: "Always"
```
nakfour: same comment for all images, adding them to opendatahub quay

ai-library/base/kustomization.yaml
```diff
@@ -0,0 +1,18 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+- crd.yaml
+- service-account.yaml
+- role.yaml
+- role-binding.yaml
+- operator.yaml
```
nakfour: @crobby  Does the order here matter? is it bottom up? or top down.?

ai-library/base/role.yaml
```diff
@@ -0,0 +1,104 @@
...
+  resources:
+  - deployments
+  - daemonsets
+  - replicasets
+  - statefulsets
+  verbs:
+  - '*'
+- apiGroups:
+  - monitoring.coreos.com
+  resources:
```
nakfour: @crobby what is this resource used for?

crobby: good question....I'll do some experimenting.  I think this came from using the operator sdk rather than something extra that we added.

crobby:
To test this, you'll need to also have Seldon installed (always a dep of ai-library).
Here's the sample CR for AI-Library:  https://gitlab.com/opendatahub/ai-library/-/blob/master/operator/deploy/crds/ailibrary_v1alpha1_ailibrary_cr.yaml
You can set most of the models there to "enabled: false" if you're tight on resources.  When you create that CR, you should start to see the SeldonDeployments for those enabled: true models show-up.

ai-library/base/operator.yaml
crobby: Yes, I'll make that happen.

ai-library/base/kustomization.yaml
crobby: I think it's top-> down.

ai-library/base/role.yaml
crobby: Yeah, looks like this comes from using the operator sdk.

ai-library/base/kustomization.yaml
```diff
@@ -0,0 +1,18 @@
...
+namespace: opendatahub
+commonLabels:
+  opendatahub.io/component: "true"
+  component.opendatahub.io/name: ailibrary
+images:
+- name: quay.io/opendatahub/ai-library-operator
```
nakfour: Do we need to specify images here? Whats the purpose, because I see the same image name in the operator.yaml file

crobby: If you change the image here, it will override what is in the other yaml files.  ie:  to update or kustomize it beyond the base yamls, you can just tweak this either here or in a overlay.

crobby: https://github.com/kubernetes-sigs/kustomize/blob/master/docs/plugins/builtins.md#field-name-images

nakfour: @crobby so you have it here as just a placement for future changes? because today the image names match

crobby: Yeah, I figured it's nice to have something there to tweak rather than needing to look up what needs to be added to tweak it.

nakfour:
@crobby I got this error
"Back-off pulling image "quay.io/opendatahub/ai-library-operator@sha256:5fbf0d49c30ae782cd6e0f75fbd5de60c01b399e60890530b4fdd68d333be162"" Error: ImagePullBackOff

crobby:
> @crobby I got this error
> "Back-off pulling image "quay.io/opendatahub/ai-library-operator@sha256:5fbf0d49c30ae782cd6e0f75fbd5de60c01b399e60890530b4fdd68d333be162"" Error: ImagePullBackOff

All set now..thanks.

nakfour:
After discussion with @crobby and @vpavlin we decided to follow these steps
1. Create an issue to fix kfctl to not wait on namespace creation of "kubeflow"  and/or "anonymous" when kfdef namespace is not "kubeflow"
2. Add the KF profile controller to the kfdef file with namespace pointing to "opendatahub". If that works, keep it there and specify it needs to be included
3. If it does not work, then add in the odh-common component with "anonymous" namespace, and keep "opendatahub" namespace in kfdef.

crobby:
> After discussion with @crobby and @vpavlin we decided to follow these steps
> 
>     1. Create an issue to fix kfctl to not wait on namespace creation of "kubeflow"  and/or "anonymous" when kfdef namespace is not "kubeflow"
> 
>     2. Add the KF profile controller to the kfdef file with namespace pointing to "opendatahub". If that works, keep it there and specify it needs to be included
> 
>     3. If it does not work, then add in the odh-common component with "anonymous" namespace, and keep "opendatahub" namespace in kfdef.

Ok, I have tried number 2 from above.  It does not work.
I have also tried number 3 and it is working, so I'll update this PR to use number 3.

crobby:
Also, when this does get merged, I'd recommend "squash and merge".

kfdef/kfctl_openshift.yaml
```diff
@@ -0,0 +1,27 @@
+
+apiVersion: kfdef.apps.kubeflow.org/v1beta1
```
nakfour: @crobby there is an extra line on top, can you please delete it?

crobby: Done

nakfour:
@crobby since depends on Seldon, can we add a not saying that in the kfdef file?

odh-common/base/kustomization.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
```
vpavlin: Can you add comonLabels here as well?

crobby: > Can you add comonLabels here as well?

ok

ai-library/base/kustomization.yaml
```diff
@@ -0,0 +1,18 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+- crd.yaml
```
vpavlin: Should we split CRDs to a separate component? e.g.

* ai-library/cluster - CRDs, potentially ClusterRoles etc. - i.e. cluster wide resources
* ai-library/operator (or something like that)

There is an issue on KF for that:https://github.com/kubeflow/manifests/issues/391

And for us it is especially important - in case we want to be enable ODH deployment by an ordinary user, we need to make sure the user can limit the list of components to things that do not require cluster admin (theoretically):)

crobby: I suppose it's better to be in-line with kf best practices

kfdef/kfctl_openshift.yaml
```diff
@@ -0,0 +1,32 @@
...
+        path: ai-library/cluster
+    name: ai-library-cluster
+  - kustomizeConfig:
+      parameters:
+      - name: namespace
+        value: opendatahub
+      repoRef:
+        name: manifests
+        path: ai-library/operator
+    # Note:  In order to utilize ai-library, you also need to have Seldon installed
```
vpavlin: Should we have seldon enabled in the KFDef then? Like this we are basically presenting a non-working KFDef

crobby: It's always been just documented as a dep in ODH.
Seldon conversion was a separate task...once we have it here, we can change the note maybe.

nakfour:
@crobby tested and verified pods are running, this was on OCP 4.3

PR #6, opened 2020-02-26
Add supserset manifests

crobby:
Adding the super set manifests to odh. For testing please follow directions here: https://gitlab.com/opendatahub/opendatahub-operator/-/blob/master/docs/deploying-superset.adoc
Skip the ODH installation parts


crobby:
This one isn't quite working yet.  Having trouble getting my vars to get replaced.

crobby:
> @crobby is this still WIP, it has AI library everywhere, copy-->paste eh ???

No, this branch was built on top of my other PR for ai-library...once that merges, this will sit on top of it cleanly and only show the superset diffs.

nakfour:
@crobby  for superset we had a lot of configurable parameters
that are not populated. https://gitlab.com/opendatahub/opendatahub-operator/-/blob/master/deploy/crds/opendatahub_v1alpha1_opendatahub_cr.yaml

nakfour:
as @crobby mentioned the default user/password to the portal is admin/admin

crobby:
@nakfour I made a tweak to how the admin password works.  Can now be specified as a param in kfctl_openshift.yaml.  The net result is that we don't need the secret.yaml at all and we can just generate the necessary secret using kustomize's secretGenerator functionality.

superset/base/kustomization.yaml
```diff
@@ -0,0 +1,19 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
```
vpavlin: Is there any specific reason why the labels are specified in each resource rather than in kustomization.yaml?

crobby: Turns out that a side-effect of "commonLabels" is that the services also wind-up with the labels and that has a negative effect when mapping a route to a service/pod.  https://github.com/kubernetes-sigs/kustomize/issues/1009  I was not able to get the metadataLabelTransformer workaround to work.

crobby: I have created a tracking issue to remind us (me) to go back and use the labelTransformer when we have a released kfctl that supports it.  https://github.com/opendatahub-io/odh-manifests/issues/7

PR #8, opened 2020-03-19
Adding kafka operator install via strimzi as well as odh kafka cluster.

crobby:


vpavlin:
Can we change the names of the subdirs? 

You have a dir `cluster` here which deploys the instance, but we agreed to follow the best practices where `cluster` is reserved for things that the admin has to deploy - which are the subscriptions.

```
mv cluster kafka
mv operator cluster
```

crobby:
> Can we change the names of the subdirs?
> 
> You have a dir `cluster` here which deploys the instance, but we agreed to follow the best practices where `cluster` is reserved for things that the admin has to deploy - which are the subscriptions.
> 
> ```
> mv cluster kafka
> mv operator cluster
> ```

sure

kfdef/kfctl_openshift.yaml
```diff
@@ -32,7 +32,20 @@ spec:
       repoRef:
         name: manifests
         path: superset
-    name: superset
+    name: superset
```
vpavlin: Can you remove trailing spaces at the end?

PR #9, opened 2020-03-22
Add radanalytics.io spark-operator

LaVLaS:
Deploys version 1.0.7 of the radanalytics spark-operator. 

You can test the deployment by attempting to create a `SparkApplication` or `SparkCluster` custom resource

```yaml
apiVersion: radanalytics.io/v1
kind: SparkApplication
metadata:
  name: my-cluster
spec:
  mainApplicationFile: local:///opt/spark/examples/jars/spark-examples_2.11-2.3.0.jar
  mainClass: org.apache.spark.examples.SparkPi
```

```yaml
apiVersion: radanalytics.io/v1
kind: SparkCluster
metadata:
  name: my-cluster
spec:
  worker:
    instances: "1"
```

### DIFF
--- a/kfdef/kfctl_openshift.yaml
+++ b/kfdef/kfctl_openshift.yaml
@@ -57,6 +57,16 @@ spec:
         name: manifests
         path: grafana/cluster
     name: grafana-cluster
+  - kustomizeConfig:
+      repoRef:
+        name: manifests
+        path: radanalyticsio/spark/cluster
+    name: radanalyticsio-cluster
+  - kustomizeConfig:
+      repoRef:
+        name: manifests
+        path: radanalyticsio/spark/operator
+    name: radanalyticsio-spark-operator
   repos:
   - name: manifests
     uri: https://github.com/opendatahub-io/odh-manifests/tarball/v0.7-branch-openshift
--- a/radanalyticsio/spark/cluster/base/clusterrole.yaml
+++ b/radanalyticsio/spark/cluster/base/clusterrole.yaml
@@ -1,0 +1,47 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: spark-operator
+rules:
+- apiGroups:
+  - radanalytics.io
+  resources:
+  - '*'
+  verbs:
+  - '*'
+- apiGroups:
+  - ''
+  resources:
+  - pods
+  - replicationcontrollers
+  - services
+  - configmaps
+  - persistentvolumeclaims
+  verbs:
+  - create
+  - delete
+  - deletecollection
+  - get
+  - list
+  - update
+  - watch
+  - patch
+- apiGroups:
+  - apiextensions.k8s.io
+  resources:
+  - customresourcedefinitions
+  verbs:
+  - list
+- apiGroups:
+  - route.openshift.io
+  resources:
+  - routes
+  verbs:
+  - create
+  - delete
+  - deletecollection
+  - get
+  - list
+  - patch
+  - update
+  - watch
--- a/radanalyticsio/spark/cluster/base/clusterrolebinding.yaml
+++ b/radanalyticsio/spark/cluster/base/clusterrolebinding.yaml
@@ -1,0 +1,12 @@
+apiVersion: rbac.authorization.k8s.io/v1beta1
+kind: ClusterRoleBinding
+metadata:
+  name: spark-operator
+roleRef:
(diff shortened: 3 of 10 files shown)


radanalyticsio/spark-operator/base/clusterrolebinding.yaml
```diff
@@ -0,0 +1,12 @@
...
+metadata:
+  name: spark-operator-edit-resources
+roleRef:
+  kind: ClusterRole
+  name: cluster-admin
+  apiGroup: ""
+subjects:
+  - kind: ServiceAccount
+    name: spark-operator
+    namespace: default
```
crobby: is "default" really the intended namespace for this?

vpavlin: Kustomize (or kfctl?) will change that to a namespace where this is being submitted - i.e. `opendatahub`, so it does not really matter what value is there, but maybe for the sake of clarity, we could use `opendatahub`

radanalyticsio/spark-operator/base/clusterrolebinding.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: rbac.authorization.k8s.io/v1beta1
+kind: ClusterRoleBinding
+metadata:
+  name: spark-operator-edit-resources
+roleRef:
+  kind: ClusterRole
+  name: cluster-admin
```
vpavlin: Is this necessary? Can we create more specific role? We should not abuse the fact that Kubeflow has cluster-admin now

vpavlin:
> spark/cluster/base (the "cluster" dir name being the convention that indicates cluster-admin activity) and then the non-cluster admin things in another dir named something like spark/cluster/.

@crobby This sentence is a bit confusing, I believe you wanted to say `spark/spark` at the end, correct?

@LaVLaS otherwise, Chad's point is valid - can you please move CRDs and ClusterRole and ClusterRoleBinding into `spark/cluster` and the deployment into `spark/spark`?

crobby:
> > spark/cluster/base (the "cluster" dir name being the convention that indicates cluster-admin activity) and then the non-cluster admin things in another dir named something like spark/cluster/.
> 
> @crobby This sentence is a bit confusing, I believe you wanted to say `spark/spark` at the end, correct?
> 
> @LaVLaS otherwise, Chad's point is valid - can you please move CRDs and ClusterRole and ClusterRoleBinding into `spark/cluster` and the deployment into `spark/spark`?

Yes, you have successfully deciphered my statement.  Thanks!

radanalyticsio/spark-operator/base/clusterrolebinding.yaml
LaVLaS: Good catch. updated

radanalyticsio/spark-operator/base/clusterrolebinding.yaml
LaVLaS: Doh. You're right. I pulled this directly from radanalytics [operator.yaml](https://github.com/radanalyticsio/spark-operator/blob/master/manifest/operator.yaml) but it definitely doesn't need cluster-admin.

LaVLaS:
@vpavlin In response to #11, would it make the JupyterHub transition easier if I added ConfigMap support as an overlay?

crobby:
> Deploys version 1.0.7 of the radanalytics spark-operator.
> 
> You can test the deployment by attempting to create a `SparkApplication` or `SparkCluster` custom resource
> 
> ```yaml
> apiVersion: radanalytics.io/v1
> kind: SparkApplication
> metadata:
>   name: my-cluster
> spec:
>   mainApplicationFile: local:///opt/spark/examples/jars/spark-examples_2.11-2.3.0.jar
>   mainClass: org.apache.spark.examples.SparkPi
> ```
> 
> ```yaml
> apiVersion: radanalytics.io/v1
> kind: SparkCluster
> metadata:
>   name: my-cluster
> spec:
>   worker:
>     instances: "1"
> ```

I think the sparkapp needs to be updated to use spark-examples_2.11-2.4.5.jar as the jar.  With that change, I was informed that Pi is roughly 3.141675708378542, which is close enough for me.

vpavlin:
@mroman-redhat is working on integration of the SparkCluster CR into jupyterhub-singleuser-profiles, so we should be good soon with that.

Having the ability to use the ConfigMap for backwards compatibility would be great, so +1 for adding it, but feel free to do it in a follow up PR

PR #10, opened 2020-03-24
Add Grafana operator subscription to opendatahub namespace

crobby:
This is the base installation of the grafana operator via subscription.  To verify that this is working, you should be able to see the operator pod running in the opendatahub namespace after you install.
There will be another PR that will deal with creating the data sources and dashboards (actually connecting to prometheus).


LaVLaS:
The OLM install of grafana is failing because there is no `OperatorGroup` object created/present in `opendatahub`

The `OperatorGroup` below worked for me and this would satisfy the requirement for any other namespaced operators being deployed in opendatahub. Since this is a unique object per namespace then I think it should be deployed as a common artifact  
```
apiVersion: operators.coreos.com/v1
kind: OperatorGroup
metadata:
  name: opendatahub
spec:
  targetNamespaces:
    - opendatahub
```

crobby:
@LaVLaS Thank you!  This makes sense...I thought I saw a flake once, but after I did a web UI install of the operator, it went away for later deployments....that webui install must have created the operatorgroup for me in the background.

nakfour:
I test this, and it works

PR #14, opened 2020-03-26
Adding grafana datasource, dashboard and instance

crobby:
Replicating the datasource/dashboard in the grafana instance from ODH classic.

In order to see any data on the included dashboard, you will need to:

1) Have strimzi-kafka installed
2) also install prometheus (from PR #16)
3) create a servicemonitor (source below) that will let prometheus scrape from the kafka services (this will be an addition to the prometheus component soon)
4) Once those are all up, you can go to your grafana route and open the "kafka overview" dashboard.  You should see data, lit up in green, for the top 4 boxes (brokers online is probably the only non-zero number unless you're doing something with kafka).

Here's the servicemonitor that needs to be created.....

apiVersion: monitoring.coreos.com/v1
kind: ServiceMonitor
metadata:
  labels:
    component.opendatahub.io/name: prometheus
    opendatahub.io/component: 'true'
    team: opendatahub
  name: odhkafkaservicemonitor
  namespace: opendatahub
spec:
  endpoints:
    - port: prometheus
  selector: {}

### DIFF
--- a/grafana/cluster/base/subscription.yaml
+++ b/grafana/cluster/base/subscription.yaml
@@ -10,3 +10,4 @@ spec:
   source: community-operators
   sourceNamespace: openshift-marketplace
   startingCSV: grafana-operator.v2.0.0
+
--- a/grafana/cluster/base/subscription.yaml
+++ b/grafana/cluster/base/subscription.yaml
@@ -10,4 +10,3 @@ spec:
   source: community-operators
   sourceNamespace: openshift-marketplace
   startingCSV: grafana-operator.v2.0.0
-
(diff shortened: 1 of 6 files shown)


vpavlin:
LGTM

crobby:
> Can you add `commonLabels`?

Sure.  I don't think commonLabels will have the negative side effect in this case.

crobby:
> > Can you add `commonLabels`?
> 
> Sure. I don't think commonLabels will have the negative side effect in this case.

This is done.  Seems to work nicely.

PR #16, opened 2020-03-27
Adding prometheus operator

nakfour:
This PR is to migrate the prometheus component. This does require the operatorgroup.yaml that was added to the Grafana PR, please make sure you manually install it. To test this
1. Install this component
kfctl build --file=kfdef/kfctl_openshift.yaml -V
kfctl apply --file=kfdef/kfctl_openshift.yaml -V
2. Go to prometheus portal from the routes and check Status-->Targets
3. Install opendatahub from the  community operator in namespace opendatahub.
4. Edit the Spark Service "spark-operator-metrics" to include name in the port
```spec:
  ports:
    - name: web
      protocol: TCP
      port: 8080
      targetPort: 8080
```
5. go back to Prometheus Status--->Targets, and re-fresh. You should see the new target for Spark. You can also search for one of the Spark variables "log4j"
6. For components that expose a prometheus endpoint, they need to name the port "web" or create a `servicemonitor` with specifics to their endpoint with a label `team: opendatahub` since this is what our Prometheus instance is looking for.


prometheus/operator/base/servicemonitor.yaml
```diff
@@ -0,0 +1,11 @@
...
+kind: ServiceMonitor
+metadata:
+  name: odhservicemonitor
+  labels:
+    team: opendatahub
+  namespace: opendatahub
+spec:
+  selector: {}
+  endpoints:
+    - port: web
```
crobby: is "web" a fairly standard prometheus port name?  It seems to me like a port named "web" would be more inclined to be a web app?

nakfour: @crobby thats what is in the example since a lot use 8080 and yes it is an http endpoint

crobby: Ok, fair enough

vpavlin:
We merged the Spark Operator PR, so I rebased here in the web ui to be able to test it easily - seems to be working fine.

I am still not sure about the name `web`  for the port, but we'll see what other services have when we do a ODH wide check for monitoring

prometheus/operator/base/prometheus.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: monitoring.coreos.com/v1
+kind: Prometheus
+metadata:
+  name: prometheus
+  labels:
+    prometheus: k8s
+  namespace: prometheus
+spec:
+  replicas: 2
```
vpavlin, line 9: Do we need 2 replicas by default?

nakfour: No we don't

kfdef/kfctl_openshift.yaml
```diff
@@ -51,7 +45,7 @@ spec:
       repoRef:
         name: manifests
         path: kafka/kafka
-    name: kafka-cluster
+    name: kafka-cluster
```
crobby, line 48: looks like maybe an unnecessary extra space showed up here

nakfour: deleted it

vpavlin:
I am going to take a look at the READMEs for the components and try to come up with consistent templates, but since we do not have it right now, this one is fine.

Looks good, merging

PR #29, opened 2020-04-17
Move clusterroles to a correct place and enable role aggregation

vpavlin:
We have missed 4 cluster roles which are now in `operator` and are not listed/deployed, but are very necessary - they enable users to actually deploy Airflow base and cluster

As part of this PR I am also adding labels which automatically aggregate these roles for any user to `view`, `edit` or `admin` default roles - i.e. any user on the OpenShift cluster will be able to deploy Airflow  cluster to their namespace if they have enough quota and `edit` or `admin` role in a namespace.

cc @VedantMahabaleshwarkar


airflow/cluster/rbac/kustomization.yaml
```diff
@@ -2,3 +2,7 @@ namePrefix: airflow-on-k8s-operator-
 resources:
 - role.yaml
 - role_binding.yaml
+- airflowbase_editor_role.yaml
+- airflowbase_viewer_role.yaml
+- airflowcluster_editor_role.yaml
+- airflowcluster_viewer_role.yaml
```
crobby, line 8: newline

PR #59, opened 2020-05-06
Setting ai library operator image version to tag v0.6

### LINKED ISSUE #60: AI Library Operator is pinned to an old version.
The AI Library operator is pinned to sha256: 293879aec3cd52959a3b670eaf69bb2ac478847dfc4e498877984e63592f0fb1 which is incorrect.  It should be grabbing the v0.6 tag of ai library operator.

crobby:
Before this, we were referring directly to an outdated sha.

To test this change, when you deploy a KfDef that has the ai-library components in it, you should see the ai-library operator running and it should be using the v0.6.

The easy way to see this is to run...

`oc get deployment ailibrary-operator -o jsonpath="{$.spec.template.spec.containers[0].image}"`

The output should be:  
quay.io/opendatahub/ai-library-operator:v0.6

This manual test has also been added to the ailibrary.sh test.

closes #60

### DIFF
--- a/ai-library/operator/base/kustomization.yaml
+++ b/ai-library/operator/base/kustomization.yaml
@@ -12,6 +12,6 @@ commonLabels:
 images:
 - name: quay.io/opendatahub/ai-library-operator
   newName: quay.io/opendatahub/ai-library-operator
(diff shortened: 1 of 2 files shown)


LaVLaS:
LGTM.

Image deployed with no issues

PR #124, opened 2020-06-18
odh-argo testing Adding an example workflow test for argo

nakfour:
This completes #95


crobby:
I think that your new resources are not making their way into the test container image.  You probably need to update the Dockerfile.

crobby:
Like it.  I think I like having all tests in one place like this....rather than having separate basictests and functionality tests.  I might go back and refactor the ai-library stuff so that each component has exatctly one file for all of its testing.

PR #128, opened 2020-06-20
Add Makefile, run container as non-root, parameterize

vpavlin:
This PR adds a Makefile

To test it, you need podman installed and be logged in into the cluster where the test should run.

```
cd tests/
make test
```

You should see the test image built, tests to run and cluster to be cleaned up

~The image now runs as non-root and is based on ubi8/go-toolset to speed up the build time (no need to install golang).~ I could not use the ubi8/go-toolset because the CI rewrites the `FROM`, so I thought I at least put things under `$HOME`, in case we figure out how to change the image in the future...but since the image is always built from master, the tests will fail since they'll look for `/peak` instead of `~/peak`

I added parameterization of Github organization and branch for image build so that you can build the image while developing the tests.

### DIFF
--- a/airflow/operator/manager/manager.yaml
+++ b/airflow/operator/manager/manager.yaml
@@ -2,7 +2,6 @@ apiVersion: apps/v1
 kind: Deployment
 metadata:
   name: controller-manager
-  namespace: system
   labels:
     control-plane: controller-manager
 spec:
--- a/tests/Dockerfile
+++ b/tests/Dockerfile
@@ -1,10 +1,15 @@
 FROM centos:centos8
 
 ARG CHROMEDRIVER_VER=84.0.4147.30
+ARG ORG=opendatahub-io
+ARG BRANCH=master
+
+ENV HOME /root
+WORKDIR /root
 
 RUN dnf install -y bc git go-toolset python3-pip unzip && \
-    git clone https://github.com/crobby/peak && \    
-    cd /peak && \
+    git clone https://github.com/crobby/peak $HOME/peak && \    
+    cd $HOME/peak && \
     git submodule update --init
 
 RUN yum -y install https://dl.fedoraproject.org/pub/epel/epel-release-latest-8.noarch.rpm &&\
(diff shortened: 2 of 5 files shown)


crobby:
Looks like the test failure is "real".  I can show you how to get a terminal in the pod inside the system at runtime if you'd like.  For me, it has helped debug this sort of thing.

vpavlin:
I think the problem is that the actual image which is run in CI is built from master, so it expects files in /peak, but the  scripts are moved to ~/peak, which is why the test fail

crobby:
Looks like it's still not picking up the value of $HOME in the CMD

crobby:
> Looks like it's still not picking up the value of $HOME in the CMD

I forced a rebuild with and it's looking like the newly built image should include both the env var and the correct CMD.  Somehow, for the previous run, it doesn't look like the image build updated, but I forced with slash test all and the new image just built looks right.  Hopefully, the run will succeed.

crobby:
I'm not understanding what is going on here:
According to the build logs, we're building a test image here...
https://storage.googleapis.com/origin-ci-test/pr-logs/pull/opendatahub-io_odh-manifests/128/pull-ci-opendatahub-io-odh-manifests-master-odh-manifests-e2e/1275140478341222400/artifacts/build-logs/odh-manifests-tests.log
That gives us the image:  registry.svc.ci.openshift.org/ci-op-g01qx414/pipeline:odh-manifests-tests  which I can pull and run locally
doing docker inspect shows $HOME being set and the CMD = $HOME/peak/installandtest.sh

but somehow, the image that actually gets ran in the test doesn't seem to do what I'd expect

crobby:
Turns out that our prow job definition still has "commands: /peak/installandtest.sh", which must be overriding whatever our CMD is.  I've got a PR in to update this:  https://github.com/openshift/release/pull/9848  I *think* it will work and once we get this one merged, we can remove the || that I have in there right now (so tests will continue to work for PRs based on the old "commands").

crobby:
> Turns out that our prow job definition still has "commands: /peak/installandtest.sh", which must be overriding whatever our CMD is. I've got a PR in to update this: [openshift/release#9848](https://github.com/openshift/release/pull/9848) I _think_ it will work and once we get this one merged, we can remove the || that I have in there right now (so tests will continue to work for PRs based on the old "commands").

Hmm...ok, my PR for "release" to update our prow job also fails.  You can see the logs there.  I think we may have to get this merged via manual testing and then submitting a PR for the ci-system to react to these changes.

tests/Dockerfile
```diff
@@ -1,35 +1,42 @@
...
-COPY functionalitytests /peak/operator-tests/odh-manifests/functionalitytests
-COPY util /peak/operator-tests/odh-manifests
-COPY resources/kfctl_openshift.yaml /kfdef/
-COPY resources/operatorsetup /peak/
-COPY resources/ai_library_cr.yaml /peak/operator-tests/odh-manifests/resources/
-COPY scripts/install.sh /peak/
-COPY scripts/installandtest.sh /peak/
-
-RUN chmod -R 777 /kfdef && \
-    mkdir -p /.kube && \
```
crobby, line 27: based on the test results from the PR to update the kube job, we might need to change permissions on /root (or something like that) to get the copying of the kubeconfig.
From the test logs:.....
```
mkdir: cannot create directory '/root/.kube': Permission denied
+ cp /var/run/secrets/ci.openshift.io/multi-stage/kubeconfig /root/.kube/config
cp: cannot create regular file '/root/.kube/config': No such file or directory
+ chmod 755 /root/.kube/config
chmod: cannot access '/root/.kube/config': No such file or directory
+ export KUBECONFIG=/root/.kube/config
+ KUBECONFIG=/root/.kube/config
+ oc new-project opendatahub
W0622 21:13:44.846228      29 loader.go:223] Config not found: /root/.kube/config
```

tests/Dockerfile
```diff
@@ -1,35 +1,42 @@
...
+    mkdir -p .kube && \
+    chmod -R 777 .kube && \
+    chmod -R 777 peak
 
 # For local testing, you can add your own kubeconfig to the image
 # Note:  Do not push the image to a public repo with your kubeconfig
 # ADD kubeconfig /root/.kube/config
 
-CMD /peak/installandtest.sh
+CMD $HOME/peak/installandtest.sh
```
crobby, line 42: Since the ci-tests seem to ignore/override this anyway, maybe we should default to something like `$HOME/peak/run.sh basictests` so the default behavior when running locally will be to just run the test suite?

tests/Makefile
```diff
@@ -0,0 +1,22 @@
...
+
+build:
+	podman build -t $(IMAGE) --build-arg ORG=$(GIT_ORG) --build-arg BRANCH=$(GIT_BRANCH) .
+
+run:
+	oc config view --flatten --minify > /tmp/tests-kubeconfig
+	podman run -e SKIP_PODS_OUTPUT=$(SKIP_PODS_OUTPUT) -it -v /tmp/tests-kubeconfig:/tmp/kubeconfig:z $(IMAGE)
+
+clean:
+	oc delete -n opendatahub kfdef opendatahub || true
```
crobby, line 17: Maybe parameterize the project name, deafulting to opendatahub is probably fine though

crobby:
Looks like the cmd from our test config is still pointed at /peak/installandtest.sh.  If I change it now, that PR against ci will fail.  Sort of a chicken and egg thingy.  Maybe add a soft link so that /peak/installandtest.sh will still find something for now until we change the test over?

tests/scripts/installandtest.sh
```diff
@@ -3,20 +3,25 @@
 set -x
 env | sort
 mkdir -p ~/.kube
+cp /tmp/kubeconfig ~/.kube/config
 cp /var/run/secrets/ci.openshift.io/multi-stage/kubeconfig ~/.kube/config
```
crobby, line 7: This line will generate an error message when ran locally.  Probably not the end of the world, but maybe add a comment or something that it will fail outside of the openshift ci system.  I don't think we want to actually go as far suppressing the error here in case there actually is a problem in the ci system with the kubeconfig.

tests/Makefile
```diff
@@ -0,0 +1,22 @@
...
+	oc delete project opendatahub || echo -e "\n\n==> If the project deletion failed, you can try to use this script to force it: https://raw.githubusercontent.com/jefferyb/useful-scripts/master/openshift/force-delete-openshift-project\n\n"
+	#Clean up openshift-operators namespace
+	oc get csv -n openshift-operators -o name | grep strimzi-cluster-operator | xargs oc delete -n openshift-operators || true
+	oc get csv -n openshift-operators -o name | grep opendatahub-operator | xargs oc delete -n openshift-operators || true
+	oc delete subscription -n openshift-operators -l peak.test.subscription=opendatahub-operator
```
crobby, line 22: Maybe delete the [mutating|validating]webhookconfigurations for seldon (maybe katib?) as well as part of the cleanup.
Also \n

vpavlin: deletes added

tests/scripts/installandtest.sh
vpavlin: I "solved" it by doing the copy from tmp and only if it fails, do the cp from /var/run...hopefully that's better

tests/Makefile
vpavlin: SInce the `opendatahub` is hardcoded elsewhere, I don't think the parameterization is necessary here, but I added the variable anyway since the value is used multiple times

airflow/operator/manager/manager.yaml
```diff
@@ -2,7 +2,6 @@ apiVersion: apps/v1
 kind: Deployment
 metadata:
   name: controller-manager
-  namespace: system
```
crobby, line 5: Ha...good find :)

PR #134, opened 2020-06-23
Converting tests to use try_until with a default timeout of 20 minutes

### LINKED ISSUE #126: Reduce potential for flake in tests
Right now, the tests are a bit flaky.  Often it's due to the ci-system being somewhat flaky in nature.  In order to be less flaky, rewrite tests wherever possible to use "try_until" with a sufficiently high timeout value to avoid false fails in situations where things are a just running slow in the ci system.  
May also be worthwhile in looking at having our tests ran on either gcp or azure where the quota is generally more available.

As part of this, also work with the test platform team on slack to let them know of the high number of flakes that we are seeing and hopefully get some support form them to reduce/eliminate flakes related purely to the ci-system itself.

crobby:
Converting tests to use try_until with a default timeout of 20 minutes to allow for platform slowness.  Also eliminating the mandatory 7 minute sleep at the start of the tests. Closes #126


crobby:
Ok.  I expect this set of tests to run, but fail due to an issue with Seldon.  We should see the AI Library functionality tests fail, but everything else should succeed.

crobby:
This should get things passing.  I have commented out the ai-library functionality testing tests which rely on Seldon.  We can turn them back on when Seldon is fixed-up.

crobby:
Once this is in and tests are passing again, I'd like to take the opportunity to update our prow config to use either azure or gcp.

PR #137, opened 2020-06-24
Fixed to subscribe to latest seldon operator 1.2

nakfour:
This fixes issue #136


nakfour:
@crobby if it is good can you pls approve?

PR #143, opened 2020-07-02
Add mixing KFDef example

vpavlin:
I have simply added the TF Job component to the new KFDef.

I am not sure about what this file's name should be - kfctl_openshift_mix soudns weird, but we have "mixing" in the roadmap, so maybe it makes sense to stick with the term. I am open to ideas:)

To test this:

* deploy ODH operator from catalog
* Use the KFDef from this PR to deploy ODH + TF Job operators
* Build an example from https://github.com/kubeflow/tf-operator/tree/master/examples/v1/dist-mnist
   
   Or you can use my pre-built image here quay.io/vpavlin/tf-dist-mnist-test:1.0 and the example manifest:

   ```
   oc apply -f https://gist.githubusercontent.com/vpavlin/466fa44adf0c6a0e81e4c4cd6abd266a/raw/3c385732d1356eb8bd7636c645d822295eea991e/tf-job-test.yaml
   ```

   You should see bunch of workers started and running and then finish

   ```
   oc describe tfjob dist-mnist-for-e2e-test
   ```
   
   should give you success in the status

### DIFF
--- a/kfdef/kfctl_openshift_mix.yaml
+++ b/kfdef/kfctl_openshift_mix.yaml
@@ -1,0 +1,133 @@
+apiVersion: kfdef.apps.kubeflow.org/v1
+kind: KfDef
+metadata:
+  name: opendatahub-mix
+  namespace: opendatahub
+spec:
+  applications:
+  - kustomizeConfig:
+      repoRef:
+        name: manifests
+        path: odh-common
+    name: odh-common
+  #Note: AI Library components requires seldon do be deployed first, make sure to place seldon before AIlibrary
+  - kustomizeConfig:
+      repoRef:
+        name: manifests
+        path: odhseldon/cluster
+    name: odhseldon
+  - kustomizeConfig:
+      repoRef:
+        name: manifests
+        path: ai-library/cluster
+    name: ai-library-cluster
+  - kustomizeConfig:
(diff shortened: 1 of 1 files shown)


vpavlin:
@crobby What do we need to do to add this file to tests?

crobby:
> @crobby What do we need to do to add this file to tests?

Good question.
I think some refactoring of installandtest.sh and install.sh might be in order here.

Right now, it sets up kfctl_openshift.yaml with updated manifests (points at the manifests from the PR) and then creates the kfdef and runs the tests.  It doesn't bother cleaning up the KfDef and just exits after the test run.  
I think we'd want to be able to loop over a set of possible KfDef files and run the tests for each (with each KfDef file being adjusted to point at the PRs manifests, of course).
We may also want to change the tests to be able to recognize whether or not a given component is installed (ie:  is present in the currently being tested KfDef) so that the test can skip itself if it's not installed (seems more likely in the future as we might be testing many different mixes)

crobby:
Do we want to merge this as-is and then add the tests (which require some refactoring to enable this) later?

vpavlin:
Yeah, I guess that makes sense..It will also help me finish the docs PR since I will be able to use real links:-)

PR #145, opened 2020-07-08
Moving strimzi/kfaka lower in the testing kfdef to speed-up tests

crobby:


crobby:
fwiw:  Tests ran about 6.5 minutes quicker and without any obvious slowdowns.

PR #148, opened 2020-07-14
Add JupyterHub Selenium tests

vpavlin:


tests/Dockerfile
```diff
@@ -1,10 +1,17 @@
...
+ARG CHROMEDRIVER_VER=83.0.4103.39
+
+RUN dnf install -y bc git go-toolset python3-pip unzip && \
     git clone https://github.com/crobby/peak && \    
     cd /peak && \
     git submodule update --init
 
+RUN yum -y install https://dl.fedoraproject.org/pub/epel/epel-release-latest-8.noarch.rpm &&\
+    echo -e "[google-chrome]\nname=google-chrome\nbaseurl=http://dl.google.com/linux/chrome/rpm/stable/x86_64\nenabled=1\ngpgcheck=1\ngpgkey=https://dl.google.com/linux/linux_signing_key.pub" > /etc/yum.repos.d/google-chrome.repo &&\
+    yum -y install google-chrome-stable-83.0.4103.116-1.x86_64 &&\
```
vpavlin, line 12: FYI - I had to do this because I need to make sure chrome and chrome driver are compatible..so when there is a new major (?) version of chrome, we'd get image build in tests failing and we'd need to update the versions to match again:( Not sure if there is  better way..

crobby:
I've got a PR to update the ci build to use centos:8  https://github.com/openshift/release/pull/10232  After that merges, the test image build for this should succeed.

tests/basictests/jupyterhub/jhtest.py
```diff
@@ -0,0 +1,283 @@
...
+            return False
+        return True
+
+
+if __name__ == "__main__":
+    jhs = JHStress()
+    jhs.run()
+    jhs.quit()
+
+    #jhs.login()
```
crobby, line 277: Are these commented out parts likely to be useful for us?

tests/basictests/jupyterhub/jhtest.py
```diff
@@ -0,0 +1,283 @@
...
+    #jhs.run_all_cells()
+    #jhs.quit()
+
+#print(driver.page_source)
+
+
```
crobby, line 283: newline, of course

crobby:
It's not finding google-chrome-stable-83.0.4103.116-1.x86_64 in the repo.   I tried it manually and can confirm that it doesn't work with the given repo setup.  I'll see if I can find another place/setup for it.

vpavlin:
They released a new version. I think for the stability we will want to have our own base image (if that is possible in the CI) which would contain the chrome and chrome driver, so that we don't have to install it on every CI build

crobby:
> They released a new version. I think for the stability we will want to have our own base image (if that is possible in the CI) which would contain the chrome and chrome driver, so that we don't have to install it on every CI build

I'll dig around and see what is or isn't possible.  I'll also see what others have done to solve this.  We can't be the first.

vpavlin:
Comments addressed

PR #159, opened 2020-07-29
Update Airflow image to 1.10.11 tag

vpavlin:
FYI @noseka1


crobby:
/test all
Looks like cluster init fail.

PR #160, opened 2020-07-31
 Mixing ODH Prometheus and Grafana with Kubeflow

nakfour:
This is part of our efforts to mix and match components between KF and ODH. Here we add prometheus and grafana to KF install with enabled service monitor for tfjobs, pytorchjobs, katib and argo. To test this install KF using ``` kfctl_openshift_kubeflow_monitoring.yaml```
And check the "Target" page in the prometheus portal. You can also run pytorchjobs,tfjobs and katib to see specific metrics. For Argo to work this depends on this PR to merge first: https://github.com/opendatahub-io/manifests/pull/49

### DIFF
--- a/kfdef/kfctl_openshift_kubeflow_monitoring.yaml
+++ b/kfdef/kfctl_openshift_kubeflow_monitoring.yaml
@@ -1,0 +1,306 @@
+# This is the config to install Kubeflow on an existing k8s cluster.
+# If the cluster already has istio, comment out the istio install part below.
+apiVersion: kfdef.apps.kubeflow.org/v1
+kind: KfDef
+metadata:
+  name: kubeflow
+  namespace: kubeflow
+spec:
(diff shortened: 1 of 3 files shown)


vpavlin:
@nakfour Can you please rebase? It needs the Grafana update to succeed in CI

crobby:
Should we add this to the tests?  Maybe cleanup after we're done testing the first KfDef and then install this KfDef and run some set of tests for components that make sense to test for this install?  Might require some tweaks to the test runner stuff, but would likely be worthwhile as we get around to more mixing.

LaVLaS:
Deployed and prometheus is scraping metrics for workflow-controller, tf-job-operator and pytorch

nakfour:
@crobby so we have tests in odh-manifest for prom and grafana. Not sure I understand where the tests need to be added

PR #178, opened 2020-08-26
Parameterize cuda build chain

vpavlin:
@LaVLaS


LaVLaS:
Initial test with default value built the cuda images successfully for `10.1`. I will test tomorrow with a parameter for `10.2`

LaVLaS:
Can we change the cuda notebook `requests` value to half of the limit for `cpu` and `memory`. This would provide user's some flexibility to start the builds if their available resources doesn't meet the limit.

vpavlin:
> Can we change the cuda notebook `requests` value to half of the limit for `cpu` and `memory`. This would provide user's some flexibility to start the builds if their available resources doesn't meet the limit.

Done

PR #194, opened 2020-09-08
Use 'oc patch' when adding htpasswd oauth during test runs

LaVLaS:
This is an update to use `oc patch` when enabling htpasswd authentication during test runs. The current implementation will overwrite the enabled oauth `identityProviders` and replace them with test `htpasswd-provider`

```
# View the oauth identityProviders that are enabled on the cluster
$ oc get oauth cluster -o yaml
```

After you running the tests with this PR you should see the identityProvider `htpasswd-provider` enabled after `test/scripts/install.sh` runs
```
identityProviders:
  - name: htpasswd-provider
    mappingMethod: claim 
    type: HTPasswd
    htpasswd:
      fileData:
        name: htpass-secret
```


crobby:
/test all
possible flake I saw on a nightly build once too

crobby:
Test seems to be failing in the jupyterhub testing somewhere (when it doesn't just flake way before that)

LaVLaS:
@crobby Do we need to be overly concerned with adding strict checking for duplicate settings in the  `install.sh` script? I could do a check to see if `htpasswd-provider` exists in `oauth` before attempting to create it but right now the script log the error but continue if it already exists

crobby:
> @crobby Do we need to be overly concerned with adding strict checking for duplicate settings in the `install.sh` script? I could do a check to see if `htpasswd-provider` exists in `oauth` before attempting to create it but right now the script log the error but continue if it already exists

As long as whatever warnings don't kill the tests (when they shouldn't kill the tests), we don't have to get too crazy.

vpavlin:
@LaVLaS you'll need to rebase when this is merged: https://github.com/opendatahub-io/odh-manifests/pull/204

PR #212, opened 2020-10-22
Move superset database init to a kubernetes job

accorvin:
Previously the superset database init ran as an init container in the
superset pod. This resulted in the database getting initialized at every
pod startup and reverting any permissions changes.

This change moves the init call into a kubernetes job that will run only
once.

### DIFF
--- a/superset/README.md
+++ b/superset/README.md
@@ -24,3 +24,10 @@ To install Superset add the following to the `kfctl` yaml file.
 ```
 
 By default the user and password to the Superset portal is admin/admin. You can change the password by changing the value of the `SUPERSET_ADMIN_PASSWORD`. To launch the portal, go to the routes in the namespace you installed Open Data Hub and click on the route with `superset` name.
+
+### Superset Database Initialization
+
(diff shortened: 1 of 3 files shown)


vpavlin:
Looks good, my only concern is that if the pod fails too many times and/or too often we might get restart backoff and get into trouble of Superset not behaving "nicely".

crobby:
/test all
Looks like the cluster spawn job in openshift-ci failed

accorvin:
I just pushed up a fix based on testing this in our stage env.

accorvin:
FYI we've deployed this change internally now and it is working well.

crobby:
I'm ok with this change, bonus points for already having it deployed this way.

crobby (approved):

crobby:
This probably needs a rebase to give the tests a chance to pass

anishasthana:
@crobby @nakfour ci passed on this PR

LaVLaS (approved):
/lgtm

crobby:
I think this needs a rebase to pick up a few changes that should get the tests passing.

nakfour:
@accorvin please rebase

PR #215, opened 2020-10-27
Add support for ARTIFACT_DIR in CI tests

LaVLaS:
The env var ARTIFACT_DIR is used in prow tests to determine where CI tests artifacts will be retrieved at the end of the run.

The environment variables, kfdef, git patch, odh pod list and operator log will be saved as artifacts for every test run.


tests/scripts/installandtest.sh
```diff
@@ -18,14 +24,15 @@ if [ -z "${SKIP_INSTALL}" ]; then
 fi
 $HOME/peak/run.sh ${TESTS_REGEX}
```
LaVLaS, line 26: @crobby I moved the pod and operator logs out of the test error check so that we always save these logs as artifacts during a test run. Can I remove `SKIP_PODS_OUTPUT` as an variable since it only appears to be used to toggle to the test log?

tests/scripts/installandtest.sh
```diff
@@ -18,14 +24,15 @@ if [ -z "${SKIP_INSTALL}" ]; then
...
 
+if [ -z "${SKIP_PODS_OUTPUT}" ]; then
```
LaVLaS, line 27: @crobby I moved the pod and operator logs out of the test error check so that we always save these logs as artifacts during a test run. Can I remove `SKIP_PODS_OUTPUT` as an variable since it only appears to be used to toggle to the test log?

tests/scripts/install.sh
```diff
@@ -26,7 +26,7 @@ if [ -z "$PULL_NUMBER" ]; then
 else
   curl -O -L https://github.com/${REPO_OWNER}/${REPO_NAME}/pull/${PULL_NUMBER}.patch
   echo "Applying followng patch:"
-  cat ${PULL_NUMBER}.patch
+  cat ${PULL_NUMBER}.patch | tee ${ARTIFACT_DIR}/github-pr-${PULL_NUMBER}.patch
```
vpavlin, line 29: Should we rather `cp` it to the `ARTIFACT_DIR` than `tee`?

tests/scripts/install.sh
```diff
@@ -61,7 +61,7 @@ if [ -z "${OPENSHIFT_USER}" ] || [ -z "${OPENSHIFT_PASS}" ]; then
 fi
 
 echo "Creating the following KfDef"
-cat ./kfctl_openshift.yaml
+cat ./kfctl_openshift.yaml | tee ${ARTIFACT_DIR}/kfctl_openshift.yaml
```
vpavlin, line 64: Same here, does `cp` make more sense?

tests/scripts/installandtest.sh
```diff
@@ -1,12 +1,18 @@
...
 set -x
-env | sort
+env | sort | tee ${ARTIFACT_DIR}/env.txt
 mkdir -p ~/.kube
 cp /tmp/kubeconfig ~/.kube/config 2> /dev/null || cp /var/run/secrets/ci.openshift.io/multi-stage/kubeconfig ~/.kube/config
 chmod 644 ~/.kube/config
 export KUBECONFIG=~/.kube/config
 
+if [ ! -d "${ARTIFACT_DIR}/screenshots" ]; then
+  ARTIFACT_SCREENSHOT_DIR="${ARTIFACT_DIR}/screenshots"
```
vpavlin, line 11: Maybe move this var out of the `if` and use it in the condition?

tests/scripts/installandtest.sh
vpavlin: Yup, I added that to be able to prevent the pod output print when running on a local cluster

tests/scripts/install.sh
LaVLaS: Is there any need to print the patch in the main test log if we are saving it as an artifact?

LaVLaS:
@crobby @vpavlin For any of the commands that use `tee`, do we actually need to include those artifacts in the main test log? I have no hard preference for `tee` vs `cp & cat` if output is necessary

tests/scripts/installandtest.sh
LaVLaS: Update. I removed `SKIP_PODS_OUTPUT`

tests/scripts/install.sh
LaVLaS: After a chat with @crobby, I have removing the output of logs/files/... that we are saving as artifacts

tests/scripts/install.sh
LaVLaS: After a chat with @crobby, I have removed the output of logs/files/... that we are saving as artifacts

tests/scripts/installandtest.sh
LaVLaS: Good call. I created and exported `ARTIFACT_SCREENSHOT_DIR` so it is accessible by all tests running in this environment

tests/scripts/installandtest.sh
crobby: This sounds good to me.  Sorry I'm way late to the party

crobby:
I'm ok with this as is and would merge it, but I'm not seeing any any "screenshots" saved anywhere in the artifacts.  Is that expected at this point?  Is there something else required to get those?  Or are those only saved on errors?

crobby:
> I'm ok with this as is and would merge it, but I'm not seeing any any "screenshots" saved anywhere in the artifacts. Is that expected at this point? Is there something else required to get those? Or are those only saved on errors?

Nevermind...I checked and see that screenshots are only for errors.

PR #219, opened 2020-10-30
Use Postgres image from registry.rh.io to avoid breakage when OpenShift removes older versions

vpavlin:


vpavlin:
@crobby Seems like Strimzi has issues on 4.6?

crobby:
@vpavlin This test isn't on 4.6 yet...we can't merge that until this PR merges

crobby:
@vpavlin:  Nightly builds are also failing with similar (the same probably) Kafka problems.  I'd be ok merging this one as it does address the issue we're having with 4.6.

Issue #222, opened 2020-11-03
Migrate Data Catalog- Hue


nakfour:
ETA: 3 wks
**Is your feature request related to a problem? Please describe.**
A clear and concise description of what the problem is. Example. Instead of [CURRENT FEATURE], I think it would be helpful to implement [EXPLANATION OF NEW FEATURE].

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

Issue #105, opened 2020-05-27
Migrate Data Catalog- Thrift Server


nakfour:
ETA: 3 weeks

nakfour:
At least get HUE migrated

PR #227, opened 2020-11-03
feat(data-catalog): Adding Hue

tumido:
Adding Hue component Data catalog.

Part of: https://github.com/opendatahub-io/odh-manifests/issues/222, https://github.com/opendatahub-io/odh-manifests/issues/105, [DATAHUB-2294](https://projects.engineering.redhat.com/browse/DATAHUB-2294)

Based on reference implementation in https://github.com/AICoE/idh-manifests/pull/27 for Internal DH.

It will need further cleanup and extraction of parts into overlays later on (storage class, externalize the database). This is expected to happen in consecutive PRs.

### DIFF
--- a/hue/README.md
+++ b/hue/README.md
@@ -1,0 +1,53 @@
+# Cloudera Hue
+
+Deploys the Cloudera Hue server allowing data exploration on Hive and S3 buckets.
+
+Cloudera Hue is expected to be deployed along any HiveServer2 type of service. In Open Data Hub a [Spark SQL Thrift Server](../thriftserver) is used. Without Thrift Server deployment, Hue won't be able to fulfil any SQL queries. However it can still serve as an S3 browser.
+
+### Folders
+
+There is one main folder in the Hue component `hue` which contains the kustomize manifests.
+
+### Installation
+
+To install Hue add the following to the `kfctl` yaml file.
+
+```yaml
+- kustomizeConfig:
+    repoRef:
+      name: manifests
+      path: hue/hue
+  name: hue
+```
+
+### Overlays
+
+Hue component provides a single overlay.
+
+#### storage-class
+
+Customizes Hue's database to use a specific `StorageClass` for PVC, see `storage_class` parameter.
+
+### Parameters
+
+There are 4 parameters exposed vie KFDef.
+
+#### storage_class
+
+Name of the storage class to be used for PVC created by Hue's database. This requires `storage-class` **overlay** to be enabled as well to work.
+
+#### hue_secret_key
+
+Set session store secret key for Hue web server.
+
+#### s3_endpoint_url
+
+HTTP endpoint exposed by your S3 object storage solution which will be made available to Hue as the default S3 filesystem location.
+
+#### s3_is_secure
+
+Specifies if HTTPS should be used as a transport protocol. Set to `true` for HTTPS and to `false` to use HTTP. Parameter is set to `true` by default.
+
+#### s3_credentials_secret
+
+Along with `s3_endpoint_url`, this parameter configures the Hue's access to S3 object storage. Setting this parameter to any name of local Openshift/Kubernetes Secret resource name would allow Hue to consume S3 credentials from it. The secret of choice must contain `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` keys. Keep in mind, in order for this value to be respected by Spark cluster properly, it must use the same credentials. If not set, credentials from [`hue-sample-s3-secret`](hue/base/hue-sample-s3-secret.yaml) will be used instead.
--- a/hue/hue/base/hive-site-xml-secret.yaml
+++ b/hue/hue/base/hive-site-xml-secret.yaml
@@ -1,0 +1,68 @@
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  name: hue-hive-site-xml
+type: Opaque
+stringData:
+  hive-site.xml: |
+    <?xml version="1.0"?>
+    <?xml-stylesheet type="text/xsl" href="configuration.xsl"?>
+    <configuration>
+      <property>
+        <name>hive.server2.transport.mode</name>
+        <value>binary</value>
+        <description>Server transport mode. binary or http.</description>
+      </property>
+
+      <property>
+        <name>hive.server2.thrift.http.port</name>
+        <value>10000</value>
+        <description>Port number when in HTTP mode.</description>
+      </property>
+
+      <property>
+        <name>fs.s3a.aws.credentials.provider</name>
+        <value>com.amazonaws.auth.EnvironmentVariableCredentialsProvider</value>
+        <description>
+          Comma-separated class names of credential provider classes which implement
+          com.amazonaws.auth.AWSCredentialsProvider.
+
+          These are loaded and queried in sequence for a valid set of credentials.
+          Each listed class must implement one of the following means of
+          construction, which are attempted in order:
+          1. a public constructor accepting java.net.URI and
+              org.apache.hadoop.conf.Configuration,
+          2. a public static method named getInstance that accepts no
+            arguments and returns an instance of
+            com.amazonaws.auth.AWSCredentialsProvider, or
+          3. a public default constructor.
+
+          Specifying org.apache.hadoop.fs.s3a.AnonymousAWSCredentialsProvider allows
+          anonymous access to a publicly accessible S3 bucket without any credentials.
+          Please note that allowing anonymous access to an S3 bucket compromises
+          security and therefore is unsuitable for most use cases. It can be useful
+          for accessing public data sets without requiring AWS credentials.
+
+          If unspecified, then the default list of credential provider classes,
+          queried in sequence, is:
+          1. org.apache.hadoop.fs.s3a.SimpleAWSCredentialsProvider:
+            Uses the values of fs.s3a.access.key and fs.s3a.secret.key.
+          2. com.amazonaws.auth.EnvironmentVariableCredentialsProvider: supports
+              configuration of AWS access key ID and secret access key in
+              environment variables named AWS_ACCESS_KEY_ID and
+              AWS_SECRET_ACCESS_KEY, as documented in the AWS SDK.
+          3. com.amazonaws.auth.InstanceProfileCredentialsProvider: supports use
+              of instance profile credentials if running in an EC2 VM.
+        </description>
+      </property>
+
+      <property>
+        <name>fs.s3a.endpoint</name>
+        <value>$(s3_endpoint_url)</value>
+        <description>AWS S3 endpoint to connect to. An up-to-date list is
+          provided in the AWS Documentation: regions and endpoints. Without this
+          property, the standard region (s3.amazonaws.com) is assumed.
+        </description>
+      </property>
+    </configuration>
--- a/hue/hue/base/hue-ini-secret.yaml
+++ b/hue/hue/base/hue-ini-secret.yaml
@@ -1,0 +1,210 @@
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  name: hue-ini
+type: Opaque
+stringData:
+  hue.ini: |
+    # Full configuration:
+    # https://github.com/cloudera/hue/blob/master/desktop/conf.dist/hue.ini
+
+    [desktop]
+      # Hide unused apps
+      app_blacklist=impala,security,jobbrowser,jobsub,pig,hbase,sqoop,zookeeper,spark,oozie,search
+      secret_key=$(hue_secret_key)
+      http_host=0.0.0.0
+      http_port=8000
+      time_zone=America/Los_Angeles
+      django_debug_mode=false
+      dev=false
+      database_logging=false
+      send_dbug_messages=false
+      http_500_debug_mode=false
+      enable_prometheus=true
+
+      [[django_admins]]
+
+      [[custom]]
+
+      [[auth]]
+
+      [[ldap]]
+        [[[users]]]
+
+        [[[groups]]]
+
+        [[[ldap_servers]]]
+
+      [[vcs]]
+
+      [[database]]
+        engine=mysql
+        host=hue-mysql.$(namespace).svc
+        port=3306
+        user=$(database_user)
+        password=$(database_password)
+        name=$(database_name)
+
+      [[session]]
+
+      [[smtp]]
+        host=localhost
+        port=25
+        user=
+        password=
+        tls=no
+
+      [[knox]]
+
+      [[kerberos]]
+
+      [[oauth]]
+
+      [[oidc]]
+
+      [[metrics]]
+
+      [[tracing]]
+
+      [[task_server]]
+
+      [[gc_accounts]]
+        [[[default]]]
+
+    [notebook]
+      [[interpreters]]
+        [[[hive]]]
+          name=Hive
+          interface=hiveserver2
+
+        [[[impala]]]
+          name=Impala
+          interface=hiveserver2
+
+        [[[sparksql]]]
+          name=SparkSql
+          interface=hiveserver2
+
+        [[[text]]]
+          name=Text
+          interface=text
+
+        [[[markdown]]]
+          name=Markdown
+          interface=text
+
+    [dashboard]
+      is_enabled=true
+      has_sql_enabled=true
+      has_report_enabled=true
+      use_gridster=true
+      has_widget_filter=false
+      has_tree_widget=false
+
+      [[engines]]
+        [[[solr]]]
+        analytics=true
+        nesting=true
+
+        [[[sql]]]
+        analytics=true
+        nesting=false
+
+    [hadoop]
+      [[hdfs_clusters]]
+        [[[default]]]
+
+      [[yarn_clusters]]
+        [[[default]]]
+
+    [beeswax]
+      hive_server_host=thriftserver.$(namespace).svc
+      hive_server_port=10000
+      hive_conf_dir=/etc/hive/conf
+      thrift_version=7
+
+      [[ssl]]
+
+    [metastore]
+      enable_new_create_table=true
+      force_hs2_metadata=false
+
+    [impala]
+      [[ssl]]
+
+    [spark]
+
+    [oozie]
+
+    [filebrowser]
+
+    [pig]
+
+    [sqoop]
+
+    [proxy]
+
+    [hbase]
+
+    [search]
+
+    [libsolr]
+
+    [indexer]
+
+    [jobsub]
+
+    [jobbrowser]
+
+    [security]
+
+    [zookeeper]
+      [[clusters]]
+        [[[default]]]
+
+    [useradmin]
+      [[password_policy]]
+
+    [liboozie]
+      oozie_url=
+
+    [aws]
+      [[aws_accounts]]
+        [[[default]]]
+          host=$(s3_endpoint_url)
+          is_secure=$(s3_is_secure)
+          calling_format=boto.s3.connection.OrdinaryCallingFormat
+          access_key_id_script=/opt/hue/aws_access_key_id.sh
+          secret_access_key_script=/opt/hue/aws_secret_access_key.sh
+
+    [azure]
+      [[azure_accounts]]
+        [[[default]]]
+
+      [[adls_clusters]]
+        [[[default]]]
+
+      [[abfs_clusters]]
+        [[[default]]]
+
+    [libsentry]
+
+    [libzookeeper]
+
+    [librdbms]
+      [[databases]]
+
+    [libsaml]
+
+    [liboauth]
+
+    [kafka]
+      [[kafka]]
+
+    [metadata]
+      [[manager]]
+      [[optimizer]]
+      [[catalog]]
+      [[navigator]]
+      [[prometheus]]
--- a/hue/hue/base/hue-mysql-pvc.yaml
+++ b/hue/hue/base/hue-mysql-pvc.yaml
@@ -1,0 +1,11 @@
+---
+kind: PersistentVolumeClaim
+apiVersion: v1
+metadata:
+  name: hue-mysql
+spec:
+  accessModes:
+    - ReadWriteOnce
+  resources:
+    requests:
+      storage: "1Gi"
--- a/hue/hue/base/hue-mysql-secret.yaml
+++ b/hue/hue/base/hue-mysql-secret.yaml
@@ -1,0 +1,10 @@
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  name: hue-mysql
+stringData:
+  database-user: datacatalog
+  database-password: datacatalog
+  database-name: datacatalog
+  database-root-password: root
--- a/hue/hue/base/kustomization.yaml
+++ b/hue/hue/base/kustomization.yaml
@@ -1,0 +1,107 @@
+---
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+  - hive-site-xml-secret.yaml
+  - hue-ini-secret.yaml
+  - hue-mysql-pvc.yaml
+  - hue-mysql-secret.yaml
+  - hue-mysql-service.yaml
+  - hue-mysql.yaml
+  - hue-mysqld-exporter-secret.yaml
+  - hue-route.yaml
+  - hue-sample-s3-secret.yaml
+  - hue-service.yaml
+  - hue.yaml
+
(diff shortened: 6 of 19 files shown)


tumido:
cc @maulikjs, @rimolive

tumido:
/retest

test failures not related to the code change

hue/hue/base/hive-site-xml-secret.yaml
```diff
@@ -0,0 +1,38 @@
...
+
+      <property>
+        <name>hive.server2.thrift.http.port</name>
+        <value>10000</value>
+        <description>Port number when in HTTP mode.</description>
+      </property>
+
+      <property>
+        <name>fs.s3a.awsAccessKeyId</name>
+        <value></value>
```
rimolive, line 26: Shouldn't this line use a variable to populate the S3 Access Key

hue/hue/base/hive-site-xml-secret.yaml
```diff
@@ -0,0 +1,38 @@
...
+      </property>
+
+      <property>
+        <name>fs.s3a.awsSecretAccessKey</name>
+        <value></value>
```
rimolive, line 31: Same as above

hue/hue/base/hue-ini-secret.yaml
```diff
@@ -0,0 +1,179 @@
...
+
+      [[vcs]]
+
+      [[database]]
+        engine=django.db.backends.mysql
+        host=hue-mysql.$(namespace).svc
+        port=3306
+        user=datacatalog
+        password=datacatalog
+        name=datacatalog
```
rimolive, line 42: These lines should be parametrized or at least generate a random password.

hue/hue/base/hue-ini-secret.yaml
```diff
@@ -0,0 +1,179 @@
...
+      [[password_policy]]
+
+    [liboozie]
+      oozie_url=
+
+    [aws]
+      [[aws_accounts]]
+        [[[default]]]
+          access_key_id=
+          secret_access_key=
```
rimolive, line 154: These lines should be populated

rimolive (changes requested):

hue/hue/base/hive-site-xml-secret.yaml
tumido: @rimolive all requested changes should be fixed now, please check again :+1:

hue/hue/base/params.env
```diff
@@ -0,0 +1,4 @@
+storage_class=
+s3_endpoint=
+s3_region=
+s3_is_secure=false
```
rimolive, line 4: It is missing the parameters for the S3 keys

hue/hue/base/hue-ini-secret.yaml
```diff
@@ -0,0 +1,179 @@
...
+      [[password_policy]]
+
+    [liboozie]
+      oozie_url=
+
+    [aws]
+      [[aws_accounts]]
+        [[[default]]]
+          access_key_id=$(s3_key_id)
+          secret_access_key=$(s3_access_key)
```
rimolive, line 154: It is better if you name the variables as `s3_access_key_id` and `s3_secret_access_key_id` as they better represent what we are storing in these variables

rimolive (changes requested):

hue/hue/base/params.env
tumido: let's discuss this at https://github.com/opendatahub-io/odh-manifests/pull/228#discussion_r521379415 it's the same thing.

hue/hue/base/hue-ini-secret.yaml
tumido: done

hue/hue/base/hue-dc.yaml
```diff
@@ -0,0 +1,70 @@
...
+          volumeMounts:
+            - name: "hue-ini"
+              mountPath: /hue/desktop/conf/
+            - name: "hive-site-xml"
+              mountPath: /etc/hive/conf/
+          livenessProbe:
+            failureThreshold: 4
+            httpGet:
+              path: /
+              port: 8000
```
vpavlin, line 46: Should the port 8000 be exposed for the container?

hue/hue/base/hue-dc.yaml
```diff
@@ -0,0 +1,70 @@
+---
+apiVersion: apps.openshift.io/v1
+kind: DeploymentConfig
```
vpavlin, line 3: Should we use `Deployment` instead of DC since (IIUC) Deployment is now preferred by OpenShift?

hue/hue/base/hue-ini-secret.yaml
```diff
@@ -0,0 +1,179 @@
...
+kind: Secret
+metadata:
+  name: hue-ini
+type: Opaque
+stringData:
+  hue.ini: |
+    [desktop]
+      # Hide unused apps
+      app_blacklist=impala,security,jobbrowser,jobsub,pig,hbase,sqoop,zookeeper,spark,oozie,search
+      secret_key=thisISaSECRET_1234
```
vpavlin, line 12: What is this `secret_key`?

hue/hue/base/hue-mysql-dc.yaml
```diff
@@ -0,0 +1,114 @@
+---
+apiVersion: apps.openshift.io/v1
+kind: DeploymentConfig
```
vpavlin, line 3: Same question - should we use `Deployment`?

hue/hue/base/hue-mysql-dc.yaml
```diff
@@ -0,0 +1,114 @@
...
+metadata:
+  name: hue-mysql
+  annotations:
+    template.alpha.openshift.io/wait-for-ready: "true"
```
vpavlin, line 7: What is this doing?

hue/hue/base/hue-dc.yaml
tumido: It would definitely simplify things, I'll look into that. The DC was chosen to mirror Internal DH's and the legacy ODH setup. Would you rather see it changed now and here or in a consecutive PR?

hue/hue/base/hue-mysql-dc.yaml
```diff
@@ -0,0 +1,114 @@
...
+  strategy:
+    type: Recreate
+  triggers:
+    - type: ImageChange
+      imageChangeParams:
+        automatic: true
+        containerNames:
+          - mysql
+        from:
+          kind: ImageStreamTag
```
vpavlin, line 18: Is this an IS coming from `openshift` namespace? Do we need more control over version?

hue/hue/base/hue-mysql-dc.yaml
```diff
@@ -0,0 +1,114 @@
...
+                secretKeyRef:
+                  name: "hue-mysql-secret"
+                  key: database-name
+            - name: MYSQL_ROOT_PASSWORD
+              valueFrom:
+                secretKeyRef:
+                  name: "hue-mysql-secret"
+                  key: database-root-password
+          resources:
+            limits:
```
vpavlin, line 97: I think we should set both limits and requests, otherwise (IIRC) if only limits is set, requests==limits

hue/hue/base/hue-mysql-pvc.yaml
```diff
@@ -0,0 +1,13 @@
+---
+kind: PersistentVolumeClaim
+apiVersion: v1
+metadata:
+  name: hue-mysql
+  annotations:
+    volume.beta.kubernetes.io/storage-class: "$(storage_class)"
```
vpavlin, line 7: I am not sure if we need this. We had it for JupyterHub because we had one deployment (MOC) which did not set default storage class and then one (not sure, maybe early IDH?) where default was not what we wanted to use.

The problem with having storage class here like this is, that you will always need to provide it - since storage class can have 3 values - `null`, ``""`, actual name...so in this case if `storage_class` is not set, it would end up being `""` and fail deployment in most cases.

You can take a look at JupyterHub - we have an overlay `storageClass` which adds the definitions and parameter if enabled.

WDYT?

hue/hue/base/hue-mysql-secret.yaml
```diff
@@ -0,0 +1,16 @@
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  name: hue-mysql-secret
+  annotations:
+    template.openshift.io/expose-username: "{.data['database-user']}"
```
vpavlin, line 7: How are these used?

hue/hue/base/kustomization.yaml
```diff
@@ -0,0 +1,133 @@
...
+  - hue-mysql-service.yaml
+  - hue-mysqld-exporter-secret.yaml
+  - hue-route.yaml
+  - hue-secret.yaml
+  - hue-service.yaml
+
+commonLabels:
+  opendatahub.io/component: "true"
+  component.opendatahub.io/name: hue
+  component.opendatahub.io/part-of: datacatalog
```
vpavlin, line 21: I like this:)

hue/hue/base/kustomization.yaml
```diff
@@ -0,0 +1,133 @@
...
+    newName: shgriffi/hue
+    newTag: test
+  - name: mysql
+    newName: openshift/mysql
+    newTag: "5.7"
+  - name: mysqld-exporter
+    newName: prom/mysqld-exporter
+
+patchesJson6902:
+  - patch: |-
```
vpavlin, line 123: Ah, you can partially disregard my comment about the image above:)

But, the problem with using images from `openshift` namespace is that versions disappear between OCP version releases - we hit it every now and then with JupyterHub DB. I would rather us to have control over when we update, then disappearing image forcing us to - can we rather pull the image from RH container catalog?

hue/hue/base/hue-dc.yaml
vpavlin: I personally am fine with having the change in this PR, up to you

hue/hue/base/hue-ini-secret.yaml
tumido: It's a django app secret key

See: https://docs.cloudera.com/documentation/enterprise/5-7-x/topics/cdh_ig_hue_config.html#topic_15_6_2__title_371

Similar to the superser's secret key https://github.com/opendatahub-io/odh-manifests/blob/58b94891d0808fd333271a0a282b018af4e2ab81/superset/base/configmap.yaml#L24

I think we should rather have it loaded from the `params.env`, correct?

hue/hue/base/hue-mysql-secret.yaml
tumido: No idea. This comes from the original implementation in the legacy ODH operator.

hue/hue/base/hue-mysql-dc.yaml
tumido: If we change this to the `Deployment` this whole section will go away.

~Also it's already controlled in the kustomization (including the namespace).~ You've found that. :)

hue/hue/base/hue-mysql-dc.yaml
tumido: @rimolive do you have any suggestion on what would be reasonable values here?

hue/hue/base/hue-mysql-pvc.yaml
tumido: We can put it into an overlay. I actually initially wanted to do that, however I kinda tried to keep the initial PR as compact as possible and not introduce unnecessary complexity. I know the storage class in data catalog is used in the Internal DH with different values for stage/prod, so I think this parameter still applies.

hue/hue/base/kustomization.yaml
tumido: :) Sure, we can change the base image.

hue/hue/base/hue-mysql-dc.yaml
tumido: Same answer. :smile: you tell me. :slightly_smiling_face: 

I have no preference. `Deployment` makes it more close to upstream kubernetes and lowers complexity of this resource manifest.

hue/hue/base/hue-mysql-dc.yaml
tumido: https://docs.openshift.com/container-platform/4.2/openshift_images/using-templates.html#templates-waiting-for-readiness_using-templates

I have no idea if it actually has any meaningful impact for ODH operator though. It comes from the initial implementation in the legacy ODH.

hue/hue/base/kustomization.yaml
tumido: It kinda mimicks the https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/ scheme :)

hue/hue/base/kustomization.yaml
tumido: removed since `Deployment`s don't support triggers anyways (changed from `DeploymentConfig`

hue/hue/base/hue-mysql-secret.yaml
tumido: removed

hue/hue/base/hue-mysql-pvc.yaml
tumido: Moved to overlay

hue/hue/base/hue-mysql-dc.yaml
tumido: I've just put some random values in there that seemed believable and sensible to me.

hue/hue/base/hue-mysql-dc.yaml
tumido: Changed to `Deployment`

hue/hue/base/hue-mysql-dc.yaml
tumido: Changed to `Deployment`

hue/hue/base/hue-ini-secret.yaml
tumido: Templated via `params.env`

hue/hue/base/hue-dc.yaml
tumido: Changed to `Deployment`

hue/hue/base/hue-dc.yaml
tumido: Fixed

tumido:
CI is broken, I'll retest it later.

hue/hue/base/kustomization.yaml
rimolive: Is this something we need to change? If not, can we mark as Resolved?

tumido: I think it can stay. :wink:

hue/hue/base/params.env
tumido: resolved via

```
s3_credentials_secret=thriftserver-sample-s3
```

hue/hue/base/hive-site-xml-secret.yaml
```diff
@@ -0,0 +1,77 @@
...
+
+      <property>
+        <name>fs.s3a.awsAccessKeyId</name>
+        <value>AWS access key ID</value>
+        <description>Enables S3 browser. Safety valve.</description>
+      </property>
+
+      <property>
+        <name>fs.s3a.awsSecretAccessKey</name>
+        <value>AWS secret key</value>
```
rimolive, line 68: It's missing the parameter to specify the AWS Secret key from params.env

hue/hue/base/hive-site-xml-secret.yaml
```diff
@@ -0,0 +1,77 @@
...
+        <description>Enables S3 browser. Safety valve.</description>
+      </property>
+
+      <property>
+        <name>fs.s3a.endpoint</name>
+        <value>AWS S3 endpoint</value>
```
rimolive, line 74: Same here.

rimolive (changes requested):

hue/hue/base/hive-site-xml-secret.yaml
tumido: This PR is WIP for a reason. Please don't review now.

tumido: Btw. The aim is to load them from environment, see above. It can't be specified here as inlined values, since that would require us traversing vars multiple times and kustomize can't do that. (a parameter defines which secret holds the s3 credentials -> loaded into vars -> based on this var we load another secret and store the creds into another variables and then template this file, that's simply not possible).

It should be doable to load it from the environment, so don't mind these values at all. I should be able to bypass it.

hue/hue/base/hive-site-xml-secret.yaml
tumido: Irrelevant, closing.

tumido:
This is also finally ready. All issues when communicating with Spark are resolved now.

Key updates:
- Huge Hue image upgrade - using 4.8.0 now from my own build, will try to get it into [ODH's quay repo for Hue](https://quay.io/repository/opendatahub/hue). Using multistage build so the image size was effectively reduced from 2.8 GiB (total image size) to 1.3 GiB
- Hue can source it's S3 credentials from environment via custom scripts and additional Hive settings - Similarly to Thrift Server, user can specify the s3 secret which holds their S3 credentials and Hue will source that.
- Simplified container's (Hue runtime and DB init job) commands
- Previously used `livelinessProbe` caused the container to be killed and restarted often because it tried to load the whole UI without any JS cache, switched over to a lightweight `is_alive` endpoint
- Can list S3 files in file browser as well as in S3 application
- Can talk to Thift Server and populate Hive meta store with external/internal tables etc.

Standalone ODH Datacatalog can be deployed via: https://gist.github.com/tumido/6b91a18954041d1f4ba13bc5eeb09d8f

Here are some screenshots:
image
image
image
image
image

tumido:
bump @accorvin @rimolive can we have this one reviewed once more?

nakfour:
@LaVLaS and @vpavlin please re-review

vpavlin:
@tumido I have just noticed the directory structure is `hue/hue/base` - any particular reason why the `hue` dir is doubled?

hue/hue/base/hue-db-init-job.yaml
```diff
@@ -0,0 +1,28 @@
+---
+apiVersion: batch/v1
```
vpavlin, line 2: Thinking about this a bit more - what happens when we upgrade and the Jub is reconciled? How do we make it run again? (IIRC Job only runs once and there is no way to retrigger it other than delete and apply again)

hue/hue/base/hue-mysql-pvc.yaml
```diff
@@ -0,0 +1,11 @@
...
+kind: PersistentVolumeClaim
+apiVersion: v1
+metadata:
+  name: hue-mysql
+spec:
+  accessModes:
+    - ReadWriteOnce
+  resources:
+    requests:
+      storage: "1Gi"
```
vpavlin, line 11: I have a very little understanding about what the DB is actually used for - do you know if 1Gi is a reasonable size?

hue/README.md
```diff
@@ -0,0 +1,47 @@
+# Cloudera Hue
```
vpavlin, line 1: It seems like this does not work without thrifserver, right? If so, can you mention it in the readme and maybe even as a comment in KFDef?

hue/hue/base/params.yaml
```diff
@@ -0,0 +1,12 @@
+---
+varReference:
+  - path: stringData/hue.ini
+    kind: Secret
+  - path: stringData/hive-site.xml
+    kind: Secret
+  - path: stringData/DATA_SOURCE_NAME
+    kind: Secret
+  - path: metadata/annotations/volume.beta.kubernetes.io/storage-class
```
vpavlin, line 9: This does not template the variable correctly..see https://github.com/opendatahub-io/odh-manifests/blob/master/jupyterhub/jupyterhub/base/params.yaml#L7

hue/hue/base/hue-mysql-pvc.yaml
maulikjs: It's used to store tables and metadata, we have been running 1 Gi in prod for some time now and it hasn't complained.

image

vpavlin (changes requested):
I looks good if the comments I added get fixed!

Thanks!

hue/hue/base/kustomization.yaml
```diff
@@ -0,0 +1,108 @@
...
+      name: hue-mysql
+    fieldref:
+      fieldpath: stringData.database-name
+
+configurations:
+  - params.yaml
+
+images:
+  - name: hue
+    newName: quay.io/tcoufal/hue
```
LaVLaS, line 101: @tumido I pushed your personal image to quay.io/opendatahub/hue, can you update the PR to reflect that image

hue/hue/base/hue-mysql-pvc.yaml
tumido: Same answer as at the Thrift server PR. I've used values based on the internal data hub team expertise. :slightly_smiling_face:

Hue stores mainly it's "user" data and preferences, (it's an old school Django app), query history, sample data cache for individual tables, data autocomplete and hive metadata cache, etc... In our experience this value is enough, though I don't really know how much lower we can go.

hue/hue/base/hue-db-init-job.yaml
tumido: Gotcha. Good point. I think we should rather have it as an init container for Hue then. WDYT? :+1:

hue/README.md
tumido: It can work without thriftserver, but you'd need to connect it to something else (yarn, Hive directly, etc) via modifying the `hue.ini`. Or, you can use it as an S3 explorer. Which is a standalone functionality. In our scenario and set up it requires Thrift server to be available. I'll explain it in the readme. :+1:

hue/hue/base/kustomization.yaml
tumido: @LaVLaS sure thing! :+1:  I'll reroute this and relieve my quay account of the Hue duty. :slightly_smiling_face:

hue/README.md
tumido: I've added a paragraph in there. WDYT? :slightly_smiling_face:

hue/hue/base/hue-db-init-job.yaml
tumido: Moved :+1:

hue/hue/base/kustomization.yaml
```diff
@@ -0,0 +1,108 @@
+---
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+  - hive-site-xml-secret.yaml
+  - hue-db-init-job.yaml
```
LaVLaS, line 6: This file is missing from the commit

tumido: I've removed it and forgot it in the kustomization, thank you! :smile: 

https://github.com/opendatahub-io/odh-manifests/pull/227#discussion_r536092305

hue/README.md
```diff
@@ -0,0 +1,49 @@
...
+
+#### storage_class
+
+Name of the storage class to be used for PVC created by Hue's database. This requires `storage-class` **overlay** to be enabled as well to work.
+
+#### hue_secret_key
+
+Set session store secret key for Hue web server.
+
+#### s3_endpoint_url
```
LaVLaS, line 43: Can you add `s3_is_secure` to the parameters in the readme? It would be required for testing this with the unsecure ceph-nano-0 endpoint

LaVLaS:
I only tested the S3 browser and it worked perfectly to view & upload files.

hue/README.md
tumido: Oh.. I've removed that variable and re-add it. And I forgot to re-add the readme section. Just a sec. Appologies.

tumido: @LaVLaS done

hue/README.md
LaVLaS: This clarification looks good to me.

crobby (approved):
This is working well for me.  I'm not a hue expert, so I'm sure I haven't exercised it much, but I approve.

LaVLaS (approved):
/lgtm

I tested the s3 browser with no issues read/writing files to ceph-nano.
I tested Hive SQL with the thriftserver component in added in #228 using the opendatahub.io data exploration [tutorial](https://opendatahub.io/docs/advanced-tutorials/data-exploration.html) without any major issues

PR #228, opened 2020-11-03
feat(data-catalog): Adding Thriftserver

tumido:
Adding Thriftserver component Data catalog.

Part of: https://github.com/opendatahub-io/odh-manifests/issues/222, https://github.com/opendatahub-io/odh-manifests/issues/105, [DATAHUB-2294](https://projects.engineering.redhat.com/browse/DATAHUB-2294)

Based on reference implementation in https://github.com/AICoE/idh-manifests/pull/29 for Internal DH.

It will need further cleanup and extraction of parts into overlays later on (storage class, externalize the database). This is expected to happen in consecutive PRs.

### DIFF
--- a/thriftserver/README.md
+++ b/thriftserver/README.md
@@ -1,0 +1,73 @@
+# Spark Thrift Server - HiveServer2
+
+Spark Thrift Server component installs HiveServer2 variant for Spark SQL - Thriftserver. It deploys the Spark SQL Thrift Server intended to expose Spark dataframes modeled as Hive tables through a JDBC connection.
+
+### Folders
+
+There is one main folder in the Thrift Server component `thriftserver` which contains the kustomize manifests.
+
+### Installation
+
+To install Thrift Server add the following to the `kfctl` yaml file.
+
+Minimal install:
+
+```yaml
+- kustomizeConfig:
+    parameters:
+      - name: spark_url
+        value: spark://spark.odh.com
+    repoRef:
+      name: manifests
+      path: thriftserver/thriftserver
+  name: thriftserver
+```
+
+Standalone install:
+
+```yaml
+- kustomizeConfig:
+    overlays:
+      - create-spark-cluster
+    parameters:
+      - name: s3_endpoint_url
+        value: s3.odh.com
+      - name: s3_credentials_secret
+        value: s3-credentials
+    repoRef:
+      name: manifests
+      path: thriftserver/thriftserver
+  name: thriftserver
+```
+
+### Overlays
+
+Thrift Server component comes with 2 overlays.
+
+#### storage-class
+
+Customizes Thrift Server to use a specific `StorageClass` for PVCs, see `storage_class` parameter.
+
+#### create-spark-cluster
+
+Requires `radanalytics/spark` component of ODH to be installed first. It provisions a minimal Spark cluster matching the Thrift Server's Spark version and connects the Thrift Server instance to it as it's master Spark cluster. This overlay modifies value of `spark_url` parameter and routes Thrift server to the Spark cluster created by this overlay only.
+
+### Parameters
+
+There are 4 parameters exposed vie KFDef.
+
+#### storage_class
+
+Name of the storage class to be used for PVCs created by Thrift Server component. This requires `storage-class` **overlay** to be enabled as well to work.
+
+#### s3_endpoint_url
+
+HTTP endpoint exposed by your S3 object storage solution which will be made available to Thrift Server as the default S3 filesystem location. In order for this value to be respected properly, the Spark cluster of choice must use the same endpoint.
+
+#### spark_url
+
+Spark cluster [`master-url`](https://spark.apache.org/docs/latest/submitting-applications.html#master-urls) in format `spark://...` which points Thrift Server to Spark cluster which it should use. This parameter value is **overriden** if `create-spark-cluster` overlay is activated. This parameter **is required** to be set if the overlay mentioned before is not used.
+
+#### s3_credentials_secret
+
+Along with `s3_endpoint_url` this parameter configures the Thrift Server's access to S3 object storage. Setting this parameter to any name of local Openshift/Kubernetes Secret resource name would allow Thift Server to consume S3 credentials from it. The secret of choice must contain `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` keys. Keep in mind, in order for this value to be respected by Spark cluster properly, it must use the same credentials. If not set, credentials from [`thriftserver-sample-s3-secret`](thriftserver/base/thriftserver-sample-s3-secret.yaml) will be used instead.
--- a/thriftserver/thriftserver/base/kustomization.yaml
+++ b/thriftserver/thriftserver/base/kustomization.yaml
@@ -1,0 +1,102 @@
+---
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+resources:
+  - thriftserver-db-exporter.yaml
+  - thriftserver-db-pvc.yaml
+  - thriftserver-db-secret.yaml
+  - thriftserver-db-service.yaml
+  - thriftserver-db.yaml
+  - thriftserver-hdfs-hive-secret.yaml
+  - thriftserver-pvc.yaml
+  - thriftserver-route.yaml
+  - thriftserver-sample-s3-secret.yaml
+  - thriftserver-server-conf-secret.yaml
+  - thriftserver-service.yaml
+  - thriftserver.yaml
+
+commonLabels:
+  opendatahub.io/component: "true"
+  component.opendatahub.io/name: thriftserver
+  component.opendatahub.io/part-of: datacatalog
+
+generatorOptions:
+  disableNameSuffixHash: true
+
+configMapGenerator:
+  - name: thriftserver-config
+    envs:
+      - params.env
+
+vars:
+  - name: namespace
+    objref:
+      kind: Service
+      apiVersion: v1
+      name: thriftserver
+    fieldref:
+      fieldpath: metadata.namespace
+  - name: storage_class
+    objref:
+      kind: ConfigMap
+      apiVersion: v1
+      name: thriftserver-config
+    fieldref:
+      fieldpath: data.storage_class
+  - name: spark_url
+    objref:
+      kind: ConfigMap
+      apiVersion: v1
+      name: thriftserver-config
+    fieldref:
+      fieldpath: data.spark_url
+  - name: s3_endpoint_url
+    objref:
+      kind: ConfigMap
+      apiVersion: v1
+      name: thriftserver-config
+    fieldref:
+      fieldpath: data.s3_endpoint_url
+  - name: s3_credentials_secret
+    objref:
+      kind: ConfigMap
+      apiVersion: v1
+      name: thriftserver-config
+    fieldref:
+      fieldpath: data.s3_credentials_secret
+  - name: database_user
+    objref:
+      kind: Secret
+      apiVersion: v1
+      name: thriftserver-db
+    fieldref:
+      fieldpath: stringData.database-user
+  - name: database_password
+    objref:
+      kind: Secret
+      apiVersion: v1
+      name: thriftserver-db
+    fieldref:
+      fieldpath: stringData.database-password
+  - name: database_name
+    objref:
+      kind: Secret
+      apiVersion: v1
+      name: thriftserver-db
+    fieldref:
+      fieldpath: stringData.database-name
+
+configurations:
+  - params.yaml
+
+images:
+  - name: spark-cluster-image
+    newName: quay.io/opendatahub/spark-cluster-image
+    newTag: 2.4.3-h2.7
+  - name: postgresql
+    newName: registry.redhat.io/rhel8/postgresql-12
+    newTag: latest
+  - name: postgres-exporter
+    newName: quay.io/internaldatahub/postgres_exporter
+    newTag: latest
--- a/thriftserver/thriftserver/base/params.env
+++ b/thriftserver/thriftserver/base/params.env
@@ -1,0 +1,4 @@
+storage_class=
+s3_endpoint_url=
+spark_url=
+s3_credentials_secret=thriftserver-sample-s3
--- a/thriftserver/thriftserver/base/params.yaml
+++ b/thriftserver/thriftserver/base/params.yaml
@@ -1,0 +1,10 @@
+---
+varReference:
+  - path: stringData/thrift-server.conf
+    kind: Secret
+  - path: stringData/DATA_SOURCE_NAME
+    kind: Secret
+  - path: metadata/annotations/volume.beta.kubernetes.io\/storage-class
+    kind: PersistentVolumeClaim
+  - path: spec/template/spec/containers[]/env[]/valueFrom/secretKeyRef/name
+    kind: Deployment
--- a/thriftserver/thriftserver/base/thriftserver-db-pvc.yaml
+++ b/thriftserver/thriftserver/base/thriftserver-db-pvc.yaml
@@ -1,0 +1,11 @@
+---
+kind: PersistentVolumeClaim
+apiVersion: v1
+metadata:
+  name: thriftserver-db
+spec:
+  accessModes:
+    - ReadWriteOnce
+  resources:
+    requests:
+      storage: 1Gi
--- a/thriftserver/thriftserver/base/thriftserver-db-service.yaml
+++ b/thriftserver/thriftserver/base/thriftserver-db-service.yaml
@@ -1,0 +1,24 @@
+---
+kind: Service
+apiVersion: v1
+metadata:
+  name: thriftserver-db
+  annotations:
+    template.openshift.io/expose-uri: |
+      postgres://{.spec.clusterIP}:{.spec.ports[?(.name=="postgres")].port}
+spec:
+  ports:
+    - name: postgres
+      protocol: TCP
+      port: 5432
+      targetPort: 5432
+    - name: exporter
+      protocol: TCP
+      port: 9187
+      targetPort: 9187
+  selector:
+    deployment: thriftserver-db
+  type: ClusterIP
+  sessionAffinity: None
+status:
+  loadBalancer: {}
--- a/thriftserver/thriftserver/base/thriftserver-pvc.yaml
+++ b/thriftserver/thriftserver/base/thriftserver-pvc.yaml
@@ -1,0 +1,11 @@
+---
+kind: PersistentVolumeClaim
+apiVersion: v1
+metadata:
+  name: thriftserver
+spec:
+  accessModes:
+    - ReadWriteOnce
+  resources:
+    requests:
+      storage: "5Gi"
--- a/thriftserver/thriftserver/overlays/create-spark-cluster/spark-cluster.yaml
+++ b/thriftserver/thriftserver/overlays/create-spark-cluster/spark-cluster.yaml
@@ -1,0 +1,22 @@
+---
+apiVersion: radanalytics.io/v1
+kind: SparkCluster
+metadata:
+  name: spark-cluster-thriftserver
+spec:
+  worker:
+    instances: "1"
+    memoryRequest: 512Mi
+    memoryLimit: 2Gi
+    cpuRequest: 200m
+    cpuLimit: "2"
+  master:
+    instances: "1"
+    memoryRequest: 512Mi
+    memoryLimit: 2Gi
+    cpuRequest: 200m
+    cpuLimit: "2"
+  customImage: spark-cluster-image
+  env:
+    - name: SPARK_METRICS_ON # Requires quay.io/radanalyticsio/openshift-spark-py36 image or compatible
+      value: prometheus
--- a/tests/basictests/thriftserver.sh
+++ b/tests/basictests/thriftserver.sh
@@ -1,0 +1,22 @@
+#!/bin/bash
+
+source $TEST_DIR/common
+
+MY_DIR=$(readlink -f `dirname "${BASH_SOURCE[0]}"`)
+
+source ${MY_DIR}/../util
+
+os::test::junit::declare_suite_start "$MY_SCRIPT"
+
+function test_thriftserver() {
+    header "Testing ODH Hue installation"
+    os::cmd::expect_success "oc project ${ODHPROJECT}"
+    os::cmd::try_until_text "oc get deployment thriftserver" "thriftserver" $odhdefaulttimeout $odhdefaultinterval
+    os::cmd::try_until_text "oc get pods -l deployment=thriftserver --field-selector='status.phase=Running' -o jsonpath='{$.items[*].metadata.name}'" "thriftserver" $odhdefaulttimeout $odhdefaultinterval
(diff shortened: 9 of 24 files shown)


tumido:
cc @maulikjs, @rimolive

tumido:
/retest

failure unrelated to PR

thriftserver/thriftserver/base/kustomization.yaml
```diff
@@ -0,0 +1,108 @@
...
+  - thriftserver-db-secret.yaml
+  - thriftserver-db-service.yaml
+  - thriftserver-dc.yaml
+  - thriftserver-hdfs-hive-secret.yaml
+  - thriftserver-pvc.yaml
+  - thriftserver-route.yaml
+  - thriftserver-server-conf-secret.yaml
+  - thriftserver-service.yaml
+
+namespace: X
```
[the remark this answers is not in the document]
tumido: Nope. That shouldn't be there. Fixed.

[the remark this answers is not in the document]
tumido: Nope. That shouldn't be there. Fixed.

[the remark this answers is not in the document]
tumido: Nope. That shouldn't be there. Fixed.

tumido:
/retest

test failures not related to the code change

thriftserver/thriftserver/base/thriftserver-secret.yaml
```diff
@@ -0,0 +1,8 @@
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  name: thriftserver-secret
+stringData:
+  s3_key_id: SECRET_KEY_ID
+  s3_access_key: SECRET_ACCESS_KEY
```
rimolive, line 8: Theses two lines should be parametrized

thriftserver/thriftserver/base/thriftserver-secret.yaml
```diff
@@ -0,0 +1,8 @@
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  name: thriftserver-secret
+stringData:
+  s3_key_id: SECRET_KEY_ID
```
rimolive, line 7: It is better if you name the variables as `s3_access_key_id` and `s3_secret_access_key_id` as they better represent what we are storing in these variables

thriftserver/thriftserver/base/params.env
```diff
@@ -0,0 +1,6 @@
+storage_class=
+s3_endpoint=
```
rimolive, line 2: It is missing the variables for the S3 keys.

rimolive (changes requested):

thriftserver/thriftserver/base/params.env
tumido: @rimolive I don't think we should be putting those in here at all. If we do that, the S3 credentials would end up in a plaintext as `parameters` in the `kfdef` and in the end they will end up in a `ConfigMap`. And that's not a safe way to treat credentials. Sensitive data should be stored in `Secrets` only. It's the same as we do for the DB credentials - we inline the defaults in the secret. See Jupyter Hub component for example. It's up to the used to override those. I know kfctl can [handle secrets](https://github.com/kubeflow/kfctl/blob/master/pkg/apis/apps/kfdef/v1/README.md#kfdefspecsecrets) and load them from the local environment (via `secrets.env`), but I don't think that applies to the operator scenario. Do you know about any sensible solution?

thriftserver/thriftserver/base/thriftserver-secret.yaml
tumido: I don't agree with that, see below.

thriftserver/thriftserver/base/thriftserver-secret.yaml
tumido: done

thriftserver/thriftserver/base/params.env
accorvin: How are we expecting ODH users to provide these values? We should have an opinionated, documented, recommendation. I think we can all agree that we don't want them in plain text somewhere, and they probably shouldn't be defined in the ODH (other than values like "CHANGEME") since they'll be unique to each deployment.

thriftserver/thriftserver/base/kustomization.yaml
```diff
@@ -0,0 +1,127 @@
...
+
+configurations:
+  - params.yaml
+
+images:
+  - name: spark-cluster-image
+    newName: quay.io/opendatahub/spark-cluster-image
+    newTag: spark24
+  - name: postgresql
+    newName: openshift/postgresql
```
vpavlin, line 111: As I described in the hue PR, we were hitting issues with missing image version between OCP releases, can we rather use RH registrY/catalog image?

thriftserver/thriftserver/base/kustomization.yaml
```diff
@@ -0,0 +1,127 @@
...
+    newTag: "9.6"
+  - name: postgres-exporter
+    newName: docker.io/wrouesnel/postgres_exporter
```
vpavlin, line 114: Any chance we can use something on quay?

thriftserver/thriftserver/base/thriftserver-db-dc.yaml
```diff
@@ -0,0 +1,101 @@
+---
+apiVersion: apps.openshift.io/v1
+kind: DeploymentConfig
```
vpavlin, line 3: Should we change to `Deployment`?

thriftserver/thriftserver/base/thriftserver-db-dc.yaml
```diff
@@ -0,0 +1,101 @@
...
+    template.alpha.openshift.io/wait-for-ready: "true"
+spec:
+  strategy:
+    type: Recreate
+  triggers:
+    - type: ImageChange
+      imageChangeParams:
+        automatic: true
+        containerNames:
+          - postgresql
```
vpavlin, line 16: Noted above, but just to be sure - can we use images from RH registry?

thriftserver/thriftserver/base/thriftserver-db-dc.yaml
```diff
@@ -0,0 +1,101 @@
...
+                secretKeyRef:
+                  name: thriftserver-db-secret
+                  key: database-password
+            - name: POSTGRESQL_DATABASE
+              valueFrom:
+                secretKeyRef:
+                  name: thriftserver-db-secret
+                  key: database-name
+          resources:
+            limits:
```
vpavlin, line 85: Can we set both requests and limits?

thriftserver/thriftserver/base/thriftserver-db-pvc.yaml
```diff
@@ -0,0 +1,13 @@
+---
+kind: PersistentVolumeClaim
+apiVersion: v1
+metadata:
+  name: thriftserver-db-pvc
+  annotations:
+    volume.beta.kubernetes.io/storage-class: "$(storage_class)"
```
vpavlin, line 7: I have described the problem with this in the #227

thriftserver/thriftserver/base/thriftserver-db-service.yaml
```diff
@@ -0,0 +1,21 @@
...
+  annotations:
+    template.openshift.io/expose-uri: |
+      postgres://{.spec.clusterIP}:{.spec.ports[?(.name=="postgresql")].port}
+spec:
+  ports:
+    - name: postgres
+      protocol: TCP
+      port: 5432
+      targetPort: 5432
+      nodePort: 0
```
vpavlin, line 15: Can we remove `nodePort`?

thriftserver/thriftserver/base/thriftserver-dc.yaml
```diff
@@ -0,0 +1,84 @@
+---
+apiVersion: apps.openshift.io/v1
+kind: DeploymentConfig
```
vpavlin, line 3: Should we switch to `Deployment`?

thriftserver/thriftserver/base/thriftserver-pvc.yaml
```diff
@@ -0,0 +1,13 @@
+---
+kind: PersistentVolumeClaim
+apiVersion: v1
+metadata:
+  name: thriftserver-pvc
+  annotations:
+    volume.beta.kubernetes.io/storage-class: "$(storage_class)"
```
vpavlin, line 7: Same as above

vpavlin:
In my opinion, we don't want to expose all those parameters. We should provide sane defaults and then ask people to follow the recommendations in http://opendatahub.io/docs/administration/installation-customization/customization.html when they need to customize things further.

thriftserver/thriftserver/base/params.env
vpavlin: @tumido I think it would be a great contribution from ODH to KF operator if we could make the secrets in KFDef actually useful - i.e. instead of pulling the values from env vars, they could be pulled from actual secrets imported to the cluster by the user. WDYT?

vpavlin: > How are we expecting ODH users to provide these values? We should have an opinionated, documented, recommendation. I think we can all agree that we don't want them in plain text somewhere, and they probably shouldn't be defined in the ODH (other than values like "CHANGEME") since they'll be unique to each deployment.

My personal approach to this is:
1. Avoid using parameters as much as possible
2. If something is unique per deployment AND is NOT a secret, put it in parameter
3. If something is unique per deployment AND IS secret, put it in secret with `CHANGME` and recommend http://opendatahub.io/docs/administration/installation-customization/customization.html
4. Is something can have a sane value which only need to be changed in specific (e.g. production) deployments, put it in CM/Secret and recommend http://opendatahub.io/docs/administration/installation-customization/customization.html

vpavlin: > I know kfctl can [handle secrets](https://github.com/kubeflow/kfctl/blob/master/pkg/apis/apps/kfdef/v1/README.md#kfdefspecsecrets) and load them from the local environment (via `secrets.env`)

FYI I have spent some time digging through the code to figure out how these secrets work and I found how/where they are loaded (https://github.com/kubeflow/kfctl/blob/master/pkg/kfconfig/loaders/v1.go#L111-L130), but I could not find a single location where it would be used:D

I also found this issue: https://github.com/kubeflow/kfctl/issues/308 which supports my hunch that there is some code missing:)

My goal was to find how it's used and propose extension where the `SecretSource` could also be an actual Secret in the cluster

vpavlin: Nevermind..found it: https://github.com/kubeflow/kfctl/blob/master/pkg/kfapp/gcp/gcp.go#L183

It is used by plugins to configure platform specific values (like the example above GCP Access Token Name).

So this does not seem like something we could use, unless we want to write an OpenShift plugin...I believe these secrets are not what we want

thriftserver/thriftserver/base/thriftserver-secret.yaml
tumido: resolved via `params.env` templating a secret name in deployments

```
s3_credentials_secret=thriftserver-sample-s3
```

thriftserver/thriftserver/base/kustomization.yaml
tumido: Resolved

thriftserver/thriftserver/base/thriftserver-db-dc.yaml
tumido: Changed to `Deployment`

thriftserver/thriftserver/base/thriftserver-db-dc.yaml
tumido: Yes we can! :+1:

thriftserver/thriftserver/base/thriftserver-db-dc.yaml
tumido: Done

thriftserver/thriftserver/base/thriftserver-db-pvc.yaml
tumido: Moved to overlay

thriftserver/thriftserver/base/thriftserver-db-service.yaml
tumido: I don't see a reason to keep it. Removed.

thriftserver/thriftserver/base/thriftserver-dc.yaml
tumido: Done :+1:

thriftserver/thriftserver/base/thriftserver-pvc.yaml
tumido: Moved to overlay.

tumido:
I've reduced the list of parameters to:

- `storage_class` - used in overlay
- `s3_endpoint` and `s3_credentials_secret` for S3 connection
- `spark_url` for connection to a Spark Cluster. This value is template if `create-spark-cluster` is used (Standalone Spark cluster is provisioned for Thriftserver via the Spark component.)
```

**Comment by tumido** [comment]
This is finally ready. More detailed update comment will be added to the Hue PR.

Key updates (additions to the legacy ODH operator):
- Thrift Server can now leverage the Spark component via `create-spark-cluster` overlay to provide standalone experience. (This overlay connects Thrift Server to the provisioned Spark cluster automatically, without need for any additional setup.)
- Uses [`2.4.3-h2.7`](https://quay.io/repository/opendatahub/spark-cluster-image?tab=tags) image I couldn't make it cooperate with Hue properly when Spark 3 (`3.0.0-h3.2`) was used, because of dated Hive or some other shady business...
- S3 credentials can be specified via a parameter - user can point Thrift server to a secret which contains it. All S3 credentials are loaded via environment varibles only, no more inlining.
- Better and faster to query `livelinesProbe` (Query the API instead of UI)
- Simplified entrypoint/args usage.
- Database schema built on Thrift server startup. In later versions of Hive (which we currently don't use) there's an additional tool to do that instead, separatelly via a Job for example, but we have no other choice for now.
- Storage class overlay
- Using latest `postgres` (RH registry image) and `postrges_exporter` (internal DH Quay repository image)

**Comment by tumido** [comment]
bump @accorvin @rimolive can we have this one reviewed once more?

File: thriftserver/thriftserver/base/kustomization.yaml (hunk 5 above)
Comment (tumido) [reply thriftserver/thriftserver/base/kustomization.yaml]: FTR, fixed :slightly_smiling_face:

File: thriftserver/README.md [hunk 14]
```diff
@@ -0,0 +1,73 @@
...
+### Installation
+
+To install Thrift Server add the following to the `kfctl` yaml file.
+
+Minimal install:
+
+```yaml
+- kustomizeConfig:
+    parameters:
+      - name: spark_url
```
vpavlin, line 18: Is `spark_url` commonly used name? I've never seen Spark connection specified by URL so far (although I only saw them in OpenShift environment where that is not easily possible due to route limitations)

vpavlin: The point being, should this be something like `spark_endpoint` to not suggest it should be a DNS name?

thriftserver/README.md
```diff
@@ -0,0 +1,73 @@
...
+
+Thrift Server component comes with 2 overlays.
+
+#### storage-class
+
+Customizes Thrift Server to use a specific `StorageClass` for PVCs, see `storage_class` parameter.
+
+#### create-spark-cluster
+
+Requires `radanalytics/spark` component of ODH to be installed first. It provisions a minimal Spark cluster matching the Thrift Server's Spark version and connects the Thrift Server instance to it as it's master Spark cluster. This overlay modifies value of `spar_url` parameter and routes Thrift server to the Spark cluster created by this overlay only.
```
vpavlin, line 53: `s/spar_url/spark_url`

thriftserver/README.md
```diff
@@ -0,0 +1,73 @@
...
+
+Name of the storage class to be used for PVCs created by Thrift Server component. This requires `storage-class` **overlay** to be enabled as well to work.
+
+#### s3_endpoint_url
+
+HTTP endpoint exposed by your S3 object storage solution which will be made available to Thrift Server as the default S3 filesystem location. In order for this value to be respected properly, the Spark cluster of choice must use the same endpoint.
+
+#### spark_url
+
+Spark cluster URL in format `spark://...` which points Thrift Server to Spark cluster which it should use. This parameter value is **overriden** if `create-spark-cluster` overlay activated. This parameter **is required** to be set if the overlay mentioned before is not used.
```
vpavlin, line 69: The example (`spark.odh.com`) above is misleading if the value is expected in form `spark://...`

vpavlin: if `create-spark-cluster` overlay *is* activated

thriftserver/thriftserver/base/params.yaml
```diff
@@ -0,0 +1,10 @@
+---
+varReference:
+  - path: stringData/thrift-server.conf
+    kind: Secret
+  - path: stringData/DATA_SOURCE_NAME
+    kind: Secret
+  - path: metadata/annotations/volume.beta.kubernetes.io/storage-class
```
vpavlin, line 7: This needs escaping for `/`

thriftserver/thriftserver/base/thriftserver-db-pvc.yaml
```diff
@@ -0,0 +1,11 @@
...
+kind: PersistentVolumeClaim
+apiVersion: v1
+metadata:
+  name: thriftserver-db
+spec:
+  accessModes:
+    - ReadWriteOnce
+  resources:
+    requests:
+      storage: 1Gi
```
vpavlin, line 11: Do we have any understanding if this is a reasonable value? (I am not sure what is actually stored in the DB)

thriftserver/thriftserver/overlays/create-spark-cluster/spark-cluster.yaml
```diff
@@ -0,0 +1,12 @@
+---
+apiVersion: radanalytics.io/v1
+kind: SparkCluster
+metadata:
+  name: spark-cluster-thriftserver
+spec:
```
vpavlin, line 6: I am not sure what the defaults are for resource requests/limits and I could not find it in the Spark Operator repo (cc @tmckayus - any ideas?). Maybe memory and cpus should be explicitly stated here?

tmckayus: @vpavlin there are no default values for memory/cpu requests and limits, the pods will run with whatever openshift defaults are in effect if they are not specified.  If the "memoryLimit/memoryRequest" and/or "cpuLimit/cpuRequest" fields are specified in the CR, they will be used to set fields in the pod spec.

(also note the deprecated "memory" and "cpu" fields which set limit and request to the same value, but the new forms should be used)

thriftserver/README.md
tumido: @vpavlin It's usually referred to as as *master url*. Endpoint itself is not sufficient, because it has to also contain the used protocol like `spark://` or `yarn://` etc. 

See this reference:
https://spark.apache.org/docs/latest/submitting-applications.html#master-urls

Maybe `spark_master_url` would work better?

thriftserver/README.md
tumido: Noticed. I'll leave this open and fix in once we decide on the naming. :+1:

thriftserver/thriftserver/base/thriftserver-db-pvc.yaml
tumido: This storage size request comes from Internal DH, where the 1Gi is sufficient. Ti stores metadata about the SQL jobs spark is tasked to perform, who are the connected clients, etc...

I have no idea how it can grow in bigger deployments that that. @maulikjs, do you think this value is reasonable?

anishasthana: I don't think we've seen issues due to the DB filling up before.

thriftserver/thriftserver/overlays/create-spark-cluster/spark-cluster.yaml
tumido: @vpavlin sure, we can state some defaults in there. Would you agree that, since this is like the minimal cluster for "demo" and "test" purposes, we can go with some reasonable minimal values which are sufficient to run the cluster without much workload expected? I think, once the user really wants to use this component, they would model the worker counts, resource requirements etc to their specific need.

tumido:
Rebased to master. Tests are flaky again. I'll retest it later.

thriftserver/README.md
vpavlin: The goal is mainly to make sure people understand what to put in that value

thriftserver/thriftserver/base/thriftserver-db-pvc.yaml
vpavlin: Ok

thriftserver/thriftserver/overlays/create-spark-cluster/spark-cluster.yaml
vpavlin: My main concern was to make sure it matches whatever is expected here https://github.com/opendatahub-io/odh-manifests/blob/3d2bba5ad161f6ffb8e5de84cca8a0e4a91f6731/thriftserver/thriftserver/base/thriftserver-server-conf-secret.yaml

thriftserver/README.md
```diff
@@ -0,0 +1,73 @@
...
+
+Standalone install:
+
+```yaml
+- kustomizeConfig:
+    overlays:
+      - create-spark-cluster
+    parameters:
+      - name: s3_endpoint_url
+        value: s3.odh.com
```
LaVLaS, line 34: is the protocol required here if we need to use unsecured `http`?

thriftserver/README.md
tumido: I've left the variable name as it and updated the documentation. WDYT?

crobby:
The tests for this actually failed:  https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/opendatahub-io_odh-manifests/228/pull-ci-opendatahub-io-odh-manifests-master-odh-manifests-e2e/1337056179376885760/artifacts/odh-manifests-e2e/e2e/container-logs/test.log

Not sure why it was reported as a success.  I'll take a look at the test wrapper.

crobby:
> The tests for this actually failed: https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/opendatahub-io_odh-manifests/228/pull-ci-opendatahub-io-odh-manifests-master-odh-manifests-e2e/1337056179376885760/artifacts/odh-manifests-e2e/e2e/container-logs/test.log
> 
> Not sure why it was reported as a success. I'll take a look at the test wrapper.

Ok, I have spotted the issue with the tests and have a PR up to fix that problem:  https://github.com/opendatahub-io/odh-manifests/pull/267
Once that merges, a rebase will be necessary (and the tests will likely report failure as it should).

LaVLaS (approved):
/lgtm

I verified this with hue PR #227 and the opendatahub.io data exploration [tutorial](https://opendatahub.io/docs/advanced-tutorials/data-exploration.html)

tumido:
Rebased :+1:

thriftserver/README.md
vpavlin: Sounds good

PR #239, opened 2020-11-09
Add Prometheus servicemonitor for Jupyterhub service

4n4nd:


anishasthana (changes requested):
@4n4nd  you will need to update  as well to link to the service monitor

prometheus/operator/base/jupyterhub-servicemonitor.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
+metadata:
+  labels:
+    app: jupyterhub
+    team: opendatahub
+  name: jupyterhubservicemonitor
```
anishasthana, line 7: /s/jupyterhubservicemonitor/jupyterhub-metrics

anishasthana (changes requested):

prometheus/operator/base/jupyterhub-servicemonitor.yaml
```diff
@@ -0,0 +1,13 @@
...
+spec:
+  endpoints:
+    - port: 8080-tcp
+  selector:
+    matchLabels:
+      app: jupyterhub
```
anishasthana, line 13: Would this result in the service monitor attempting to scrape itself or something due to the labels?

prometheus/operator/base/jupyterhub-servicemonitor.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
+metadata:
+  labels:
+    app: jupyterhub
```
anishasthana, line 5: Would changing the name to odh-jupyterhub be safer?

prometheus/operator/base/jupyterhub-servicemonitor.yaml
4n4nd: it looks for services with that label

prometheus/operator/base/jupyterhub-servicemonitor.yaml
4n4nd: I think jupyterhub is better since all other JH components have this label

prometheus/operator/base/jupyterhub-servicemonitor.yaml
4n4nd: updated :+1:

prometheus/operator/base/jupyterhub-servicemonitor.yaml
anishasthana: Fine by me. I'll defer to @nakfour

vpavlin: The label is bound to the `team` right? In that case, I think we're fine

nakfour: @anishasthana is this supposed to scrape metrics from KF jupyter ? If not, we need to say it is odh-jupyterhub. I know KF does not have jupyterhub but there is always confusion between what odh has and what KF has.

4n4nd: I think a lot of ODH [JH components](https://github.com/opendatahub-io/odh-manifests/blob/master/jupyterhub/jupyterhub/base/jupyterhub-dc.yaml#L4-L5) have this label. 
Are we planning to rename them to odh-jupyterhub?

vpavlin: Not particular plan, we have other labels too, if you want to be more specific `component.opendatahub.io/name: jupyterhub`

4n4nd: I think we should keep it as app=jypyterhub rn and create a new PR to update labels for all the JH components.

nakfour: ok thats fine, please open the issue to change all labels to jupyterhub

nakfour (approved):

PR #250, opened 2020-11-24
feat(ceph): Export ceph credentials into a secret

tumido:
This PR adds an exporter sidecar to the `ceph-nano-0` pod. This exporter waits for ceph to start and then produces a secret containing the S3 credentials in a format consumable from data catalog and other services.

What is going on in this PR:
1. New `initContainer` which populates the shared directory between the `ceph-nano` and `exporter` containers. Ceph-nano stores the user profile in `/opt/ceph-container/tmp` directory. This dir already contains a file, which is baked into the image. And since Ceph expect this file to be there, we need to copy it over to the shared directory (prepopulate it with the expected content)
2. Additional sidecar container running the Openshift Client image (can be replaced by a bare curl querying REST API, but this seems cleaner to me). This sidecar uses the service account token and the user details from ceph to parse out S3 credentials. Then, it populates/updates a secret called `ceph-nano-credentials` in the namespace.
3. In order for the service account to have access to the secrets, I had to add a role for this.

This makes ceph much more useful, since the credentials can be directly mounted onto other pods, like the data catalog.

### DIFF
--- a/ceph/README.md
+++ b/ceph/README.md
@@ -31,19 +31,19 @@ To install Ceph Object Storage add the following to the `kfctl` yaml file.
 
 * Deployment of `ceph-nano` will create a route to a `ceph-nano-dashboard` that provides a S3 web portal for the in pod object storage.
 
-* The ACCESS_KEY and SECRET_KEY created for this deployment can be retrieved from the `ceph-nano-0` pod under `/nano_user_details`
+* The ACCESS_KEY and SECRET_KEY created for this deployment can be found in `ceph-nano-credentials` secret.
   ```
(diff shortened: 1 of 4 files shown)


tumido:
cc @LaVLaS

tumido:
hola team, can we get this reviewed, or at least approved for tests? @LaVLaS @vpavlin

crobby:
I like this a lot.  Will give it a try.  If the tests fail, you'll probably need to rebase to pick up https://github.com/opendatahub-io/odh-manifests/pull/256

nakfour:
@nakfour @vpavlin pls review

LaVLaS (approved):
/lgtm

PR #264, opened 2020-12-09
build tensorflow-gpu with cuda and minimal-notebook

harshad16:
build tensorflow-gpu with cuda and minimal-notebook

### DIFF
--- a/jupyterhub/notebook-images/overlays/cuda/tensorflow-gpu-notebook.yaml
+++ b/jupyterhub/notebook-images/overlays/cuda/tensorflow-gpu-notebook.yaml
@@ -34,12 +34,12 @@ items:
         from:
           kind: ImageStreamTag
           name: $(cuda_version)-cuda-python-36-ubi7:latest
+        env:
+        - name: ENABLE_MICROPIPENV
+          value: "1"
+        - name: UPGRADE_PIP_TO_LATEST
+          value: "1"
       type: Source
-      env:
-      - name: ENABLE_MICROPIPENV
-        value: "1"
-      - name: UPGRADE_PIP_TO_LATEST
-        value: "1"
     triggers:
     - type: ConfigChange
     - type: ImageChange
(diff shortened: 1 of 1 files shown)


jupyterhub/notebook-images/overlays/cuda/tensorflow-gpu-notebook.yaml
```diff
@@ -23,6 +23,10 @@ items:
         name: s2i-minimal-notebook-gpu:3.6
     resources:
       limits:
```
LaVLaS, line 25: I think we need to parameterize these "default" resource values so that it can be modified by the user in the kfdef. I think we originally had these hardcoded in the old operator but user's were reporting issues with failed builds when we set the limits which is why we removed the `limits`

jupyterhub/notebook-images/overlays/cuda/tensorflow-gpu-notebook.yaml
```diff
@@ -66,11 +70,15 @@ items:
         name: s2i-tensorflow-notebook-gpu:3.6
     resources:
       limits:
```
LaVLaS, line 72: See previous comment about parameterizing these resource values

LaVLaS (changes requested):

jupyterhub/notebook-images/overlays/cuda/tensorflow-gpu-notebook.yaml
vpavlin: This image is fine - it is the same as building the s2i-minimal-notebook image, so even these limits are probably unnecessarily high.

For the image below (s2i-tensorflow-gpu-notebook), I'd vote for leaving the limits there as they are set now - min 2 Gi, max 4 Gi - it "should" be enough. We can parameterize later if we see people reporting issues - WDYT @LaVLaS

harshad16: I understand the point, I will remove the limit from this one and leave the s2i-tensorflow-gpu-notebook one. 
 I had added them based on the build I tested in the cluster.

LaVLaS: We had an offline discussion and based on your experience with gpu notebook builds, there is no current need to add parameters for these values. We can revisit this in the event that a user has a need to modify these values.

jupyterhub/notebook-images/overlays/cuda/tensorflow-gpu-notebook.yaml
LaVLaS: See previous https://github.com/opendatahub-io/odh-manifests/pull/264#discussion_r559733780

LaVLaS (approved):

nakfour (approved):

vasukulkarni:
Can we also enforce adding some basic unit tests under tests/ folder ? thanks!

PR #268, opened 2020-12-14
Update odh-dashboard deployment imagePullPolicy to "Always"

cfchase:
Added `imagePullPolicy: Always` so the deployment would pick up newer image marked with the `v0.9` tag


LaVLaS (approved):
/lgtm

PR #271, opened 2020-12-17
[Monitoring] Replace jupyterhub servicemonitor with one servicemonitor for odh components

4n4nd:
This will remove the single servicemonitor for Jupyterhub 
Update odh service monitor to monitor endpoints for all the odh components that have metrics enabled.


anishasthana (changes requested):
We should also remove https://github.com/opendatahub-io/odh-manifests/blob/master/prometheus/operator/base/servicemonitor.yaml then. I believe this is the one for Argo.

We would also need to update https://github.com/opendatahub-io/odh-manifests/blob/master/prometheus/operator/base/kustomization.yaml.

anishasthana:
We should also discuss with the actual monitoring team to confirm this isn't an abuse of service monitors :-). If it isn't, I'm a fan of the PR.

anishasthana:
Hmm, just had a thought, what if in an ODH deployment we only deploy one or two components. Won't the service-monitor generated scrape configs tell prometheus to attempt to scrape from endpoints that don't exist?

4n4nd:
> Hmm, just had a thought, what if in an ODH deployment we only deploy one or two components. Won't the service-monitor generated scrape configs tell prometheus to attempt to scrape from endpoints that don't exist?

that should not happen because as far as I know, it generates configuration only for endpoints that are discovered

crobby:
@anishasthana @4n4nd Are we all set with the changes requested here?  Seems to be working ok.  Are we good to get this merged?

4n4nd:
> @anishasthana @4n4nd Are we all set with the changes requested here? Seems to be working ok. Are we good to get this merged?

@crobby I never got a response from the monitoring team, but I think this is the best way to go

anishasthana (approved):

crobby:
> > @anishasthana @4n4nd Are we all set with the changes requested here? Seems to be working ok. Are we good to get this merged?
> 
> @crobby I never got a response from the monitoring team, but I think this is the best way to go

Sounds good.  I'll let @nakfour weigh-in, but I think it should be good to go.

LaVLaS (approved):
/lgtm

4n4nd:
@crobby @anishasthana I am not really sure what's failing with [the CI](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/opendatahub-io_odh-manifests/271/pull-ci-opendatahub-io-odh-manifests-master-odh-manifests-e2e/1349006827961978880).

crobby:
/test odh-manifests-e2e
Could be an actual ci issue going on with their system.  Giving it one more try.

crobby:
I'm still not able to explain why *this* job is the only one showing this problem.  We had nightly runs that went through and didn't exhibit this behavior.  Since I can't watch this run "live" in the ci system, I may submit another equal PR so that I can watch the progression of the tests and maybe sort out what is happening.

crobby:
> I'm still not able to explain why _this_ job is the only one showing this problem. We had nightly runs that went through and didn't exhibit this behavior. Since I can't watch this run "live" in the ci system, I may submit another equal PR so that I can watch the progression of the tests and maybe sort out what is happening.

Oh wait....I just noticed the same thing happened on https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/opendatahub-io_odh-manifests/272/pull-ci-opendatahub-io-odh-manifests-master-odh-manifests-e2e/1349103123322376192  Maybe the problem is just happening for PR-related runs.

crobby:
Interestingly enough, my dummy PR test run:  https://github.com/opendatahub-io/odh-manifests/pull/278 did succeed.  Makes me think that I'd like to watch this test run live as it happens in the ci.  If I rekick this now, @4n4nd can you login to the openshift-ci system (ping me and I can help you with that) and grant me (crobby since it goes by github user id) access to the project inside openshift-ci so I can babysit the run?

Issue #15, opened 2020-03-27
Offline/Disconnected install


vpavlin:
https://github.com/kubeflow/kfctl/pull/226

walidshaari:
Any update on that issue, or how to workaround for the time being?

vpavlin:
Hi, not yet, it is planned for the release 0.9, so I will start working on this soon

walidshaari:
let me know if you would like me to test

vpavlin:
kfctl has a command to replace the image url prefix

code: https://github.com/kubeflow/kfctl/blob/master/cmd/kfctl/cmd/set-image-name.go
doc: https://github.com/kubeflow/kfctl/blob/master/cmd/kfctl/README.md#set-image-name

We should investigate how to get this working inside the operator. Then we can add something lik `registry:` to KFDef and allow the operator to automatically change all the images on the fly

carlmes:
A small request for offline / disconnected installs - can we also change the image references to use SHA instead of version number, for example in [jupyterhub-img-imagestream.yaml](https://github.com/opendatahub-io/odh-manifests/blob/6a186a39a1f340efea77e704f6dcbe658e2341f3/jupyterhub/jupyterhub/base/jupyterhub-img-imagestream.yaml#L14
) instead of `quay.io/odh-jupyterhub/jupyterhub-img:v0.1.4` we could either have the ability to override the postfix, or perhaps modify the source to use SHA such as instead of `quay.io/odh-jupyterhub/jupyterhub-img:v0.1.4`, the source contains `quay.io/odh-jupyterhub/jupyterhub-img@sha256:a9062538fdca6913d803bb4a9500f4fd303d022ef07a703e967e87248ad6a4c5`?

The reason for this request is that a number of proxy/mirror registries within the disconnected network utilize [Image Content Source Policy](https://docs.openshift.com/container-platform/4.6/rest_api/operator_apis/imagecontentsourcepolicy-operator-openshift-io-v1alpha1.html) requiring the images to be referenced by SHA.

vpavlin:
Attempting to PoC one approach for JupyterHub

1. Add a `parameter` `registry` and use it in all image references (DCs, ISs, CMs..)
2. Add overlay `disconnected` if necessary
3. Provide `kfdef/kfctl_openshift_disconnected.yaml` which would show how to "define" disconnected instance

[...]

PR #282, opened 2021-01-18
WIP: Add registry parameters for JH to support disconnected install

vpavlin:
This implements part of #15.

Specifically it adds parameter `registry` and `registry_db` for JH components which allow users to specify the registry where the images for the component should be pulled from (by default `quay.io` and `registry.redhat.io`)

PTAL @carlmes

### DIFF
--- a/jupyterhub/README.md
+++ b/jupyterhub/README.md
@@ -11,7 +11,7 @@ Contains deployment manifests for JupyterHub instance.
 
 ### Parameters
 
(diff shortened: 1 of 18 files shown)


vpavlin:
There is a new approach we are pursuing, closing.

shalberd:
> There is a new approach we are pursuing, closing.

@vpavlin What is that new approach? I have a customer with disconnected openshift installs and private registries. How to handle private registry image references for all the components (Jupyter, Kubeflow, Airflow ...)  ?

PR #208, opened 2020-10-19
Update odhargo component to v2.12.5

tumido:
**Merge together with https://github.com/opendatahub-io/odh-dashboard/pull/37 please**

Argo is moving forward quite fast and ODH should keep up with the progress. This PR implements 2 enhancements to the current Argo in ODH:

- Bump the manifests to the 2.12.5 release
- Manifests are generated via #289 script

That required me to change how the manifest files are structured in odhargo though - all custom ODH manifests and modifications are now static, while all the resources which are built via kustomize from the upstream are built into specific files:

For `cluster/base` those files are:
- `crds.yaml`
- `clusterroles.yaml`

For `odhargo/base` it's a single file: `namespace-install.yaml` that is very similar to the upstream's [`namespace-install.yaml`](https://github.com/argoproj/argo/blob/master/manifests/enamespace-install.yaml) but it lacks CRDs.

Target version used by the generator script is the latest release, fetched during the script execution.

One last modification to the previous behavior is that Argo can now use the `default` service account (which is used by default by argo anyways) instead of creating a custom one in and for ODH only.

PR #289, opened 2021-01-21
feat: Add Argo updater script and maintainer docs

tumido:
Follow up to: https://github.com/opendatahub-io/odh-manifests/pull/208

This PR is separating the updater script into a new PR so we can have more fun with that bit and we don't block the actual manifests from merging. 

Context: https://github.com/opendatahub-io/odh-manifests/pull/208#discussion_r558079323


nakfour:
this looks good  @tumido /lgtm

PR #299, opened 2021-02-01
Update peak operator file to include odh-manifests repo url

LaVLaS:
The `operatorsetup` file used in the peak tests had `nil` in the argument position that peak references for `git clone <REPO_URL>`.  

This replaces `nil` with the url for `odh-manifests` repo and should resolve the error below in the `manifests-e2e` tests

```
+ ./setup.sh -t /root/peak/operatorsetup
++++++++++++++ Processing entry for operator opendatahub-operator ++++++++++++++
Cloning test repository for opendatahub-operator
git clone nil /root/peak/operator-tests/opendatahub-operator
fatal: repository 'nil' does not exist
```


crobby:
probably needs a rebase to pick-up the airflow URL change

crobby:
I feel like the nil was intentional at one point, but I can't quite remember.  I think the original intent was something like we didn't need to clone the odh-manifests repo since it was already present in the image at test-time (I think that param is the test repo to clone).  But, having it clone the operator repo shouldn't be a collision of any sort, so I think this is fine....plus it seems to work fine.

PR #305, opened 2021-02-08
Simplify test for dashboard pods, hopefully eliminating flake

crobby:
Previously, the dashboard tests might fail if there was just one pod up and running instead of the expected 2.  This change should just repeat the call until there are exactly 2 pods up and running.


anishasthana (approved):
/lgtm

PR #307, opened 2021-02-10
Use "Argo Workflows" instead of "Argo"

sbose78:
.. to be more explicit about which Argo project is being used.


crobby (approved):
Thanks

PR #315, opened 2021-02-12
[Jupyterhub] exclude /metrics endpoint from auth proxy protection

### LINKED ISSUE #314: [JH] update to new JH version breaks metrics scraping
After the recent jupyterhub update, the `/metrics` enpoint is protected behind the auth proxy by default. We need to manually change the configuration to make `/metrics` endpoint scrapable.

We made this change in the MOC operate-first environment (PR: https://github.com/operate-first/apps/pull/255)

4n4nd:
Closes #314

### DIFF
--- a/jupyterhub/jupyterhub/base/jupyterhub-configmap.yaml
+++ b/jupyterhub/jupyterhub/base/jupyterhub-configmap.yaml
@@ -5,7 +5,7 @@ metadata:
     app: jupyterhub
   name: jupyterhub-cfg
 data:
-  jupyterhub_config.py: ""
+  jupyterhub_config.py: "c.JupyterHub.authenticate_prometheus = False;"
   jupyterhub_admins: "admin"
   gpu_mode: ""
   singleuser_pvc_size: 2Gi


vpavlin:
@4n4nd I dontT think this is a good approach to solve this. 

A question: Is the `/metrics` publicly accessible when we do this? I.e. Can anyone with the knowledge of the route see the metrics? Would it make more sense to take a look if we can make it so that we do not expose metrics on the route and keep it just cluster internal?

A comment: If we agree that this is a good approach, we should make the change here: https://github.com/opendatahub-io/jupyterhub-odh/blob/master/.jupyter/jupyterhub_config.py

4n4nd:
>  Is the `/metrics` publicly accessible when we do this? I.e. Can anyone with the knowledge of the route see the metrics?

Yes, it does

wseaton:
@vpavlin @4n4nd there's a secure way to do this by generating a service account secret and adding it to the Prometheus config.

```yaml
- bearerTokenSecret:
  key: token
  name: jupyterhub-prometheus-secret
port: 8080-tcp # Jupyterhub
path: "/metrics"
```
Arbitrary tokens for services can be added to the jupyterhub_config.py (would just have to expose the secret to the pod as an env var or file:

```py
c.JupyterHub.services = [{'name': 'prometheus', 'api_token': 'super-secret-token'},]
```

anishasthana:
cc @4n4nd @maulikjs

4n4nd:
@wseaton's PR(https://github.com/opendatahub-io/odh-manifests/pull/330) implements a better solution, so I can close this one

PR #316, opened 2021-02-16
Add JupyterHub dashboards

hemajv:
As part of the **OperateFirst SRE** efforts and in collaboration with the **Internal Data Hub** team, we have created the following Grafana dashboards for the JupyterHub application, which we think would be suitable basic dashboard templates to have as part of ODH:

1. [JupyterHub Monitoring dashboard](https://grafana-route-opf-monitoring.apps.cnv.massopen.cloud/d/BfSK2f1Mz/jupyterhub-sli-slo?orgId=1) - This dashboard visualizes some of the metrics useful for monitoring the JupyterHub service and is useful from an operating perspective
2. [JupyterHub Usage dashboard](https://grafana-route-opf-monitoring.apps.cnv.massopen.cloud/d/YuYfkHYMk/jupyterhub-usage?orgId=1) - This dashboard visualizes user activity on JupyterHub such as CPU, memory and PVC usage and is useful for users to track their individual JupyterHub resources

(related to https://github.com/operate-first/SRE/issues/66)

### DIFF
--- a/grafana/grafana/base/jupyterhub-sli-dashboard.yaml
+++ b/grafana/grafana/base/jupyterhub-sli-dashboard.yaml
@@ -1,0 +1,10 @@
+apiVersion: integreatly.org/v1alpha1
+kind: GrafanaDashboard
+metadata:
+  name: odh-jupyterhub-sli
+  labels:
+    app: grafana
+spec:
+  configMapRef:
+    name: jh-sli-configmap
+    key: jupyterhub-sli-slo.json
--- a/grafana/grafana/base/jupyterhub-usage-dashboard.yaml
+++ b/grafana/grafana/base/jupyterhub-usage-dashboard.yaml
@@ -1,0 +1,10 @@
+apiVersion: integreatly.org/v1alpha1
+kind: GrafanaDashboard
+metadata:
+  name: odh-jupyterhub-usage
+  labels:
+    app: grafana
+spec:
+  configMapRef:
+    name: jh-usage-configmap
+    key: jupyterhub-usage.json
--- a/grafana/grafana/base/kustomization.yaml
+++ b/grafana/grafana/base/kustomization.yaml
@@ -5,8 +5,19 @@ namespace: openshift-operators
 commonLabels:
   opendatahub.io/component: "true"
   component.opendatahub.io/name: grafana
+generatorOptions:
+ disableNameSuffixHash: true
+configMapGenerator:
+- name: jh-sli-configmap
+  files:
+  - dashboards/jupyterhub-sli-slo.json
+- name: jh-usage-configmap
+  files:
+  - dashboards/jupyterhub-usage.json
 resources:
-- datasource.yaml
-- dashboard.yaml
 - argo-dashboard.yaml
+- dashboard.yaml
+- datasource.yaml
 - grafana.yaml
+- jupyterhub-sli-dashboard.yaml
+- jupyterhub-usage-dashboard.yaml
--- a/prometheus/operator/base/cluster-metrics-servicemonitor.yaml
+++ b/prometheus/operator/base/cluster-metrics-servicemonitor.yaml
@@ -1,0 +1,41 @@
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
+metadata:
+  name: cluster-monitor-federation
+  labels:
+    monitor-component: cluster-resources
+    team: opendatahub
+spec:
+  endpoints:
+    - bearerTokenSecret:
+        key: token
+        name: odh.prometheus-k8s
+      honorLabels: true
+      params:
+        'match[]':
+          - '{__name__= "container_cpu_cfs_throttled_seconds_total"}'
+          - '{__name__= "container_cpu_usage_seconds_total"}'
+          - '{__name__= "container_memory_working_set_bytes"}'
+          - '{__name__= "container_memory_rss"}'
+          - '{__name__= "kubelet_volume_stats_used_bytes"}'
+          - '{__name__= "kubelet_volume_stats_capacity_bytes"}'
+          - '{__name__= "kube_pod_container_status_restarts_total"}'
+          - '{__name__= "kube_pod_container_status_terminated_reason"}'
+          - '{__name__= "kube_pod_container_status_waiting_reason"}'
+          - '{__name__= "kube_pod_container_resource_requests_cpu_cores"}'
+          - '{__name__= "kube_pod_container_resource_limits_cpu_cores"}'
+          - '{__name__= "kube_pod_container_resource_requests_memory_bytes"}'
+          - '{__name__= "kube_pod_container_resource_limits_memory_bytes"}'
+      path: /federate
+      port: web
+      scheme: https
+      tlsConfig:
+        insecureSkipVerify: true
+      scrapeTimeout: 10s
+      interval: 30s
+  namespaceSelector:
+    matchNames:
+      - openshift-monitoring
+  selector:
+    matchLabels:
+      prometheus: k8s
--- a/grafana/grafana/base/dashboards/jupyterhub-sli-slo.json
+++ b/grafana/grafana/base/dashboards/jupyterhub-sli-slo.json
@@ -1,0 +1,1570 @@
+{
+  "annotations": {
+    "list": [
+      {
+        "builtIn": 1,
+        "datasource": "-- Grafana --",
+        "enable": true,
+        "hide": true,
+        "iconColor": "rgba(0, 211, 255, 1)",
+        "name": "Annotations & Alerts",
+        "type": "dashboard"
+      }
+    ]
+  },
+  "editable": true,
+  "gnetId": null,
+  "graphTooltip": 0,
+  "id": 13,
+  "iteration": 1613488593726,
+  "links": [],
+  "panels": [
+    {
+      "datasource": null,
+      "fieldConfig": {
+        "defaults": {
+          "custom": {}
+        },
+        "overrides": []
+      },
+      "gridPos": {
+        "h": 8,
+        "w": 24,
+        "x": 0,
+        "y": 0
+      },
+      "id": 40,
+      "options": {
+        "content": "# JupyterHub SLI Dashboard\n\nThis dashboard helps visualize the SLIs (Service Level Indicators) important for monitoring the JupyterHub service.\n\nThe main SLIs tracked are:\n* Availability \n* Latency\n* Saturation\n* Error",
+        "mode": "markdown"
+      },
+      "pluginVersion": "7.1.0",
+      "timeFrom": null,
+      "timeShift": null,
+      "title": "JupyterHub SLI Dashboard",
+      "type": "text"
+    },
+    {
+      "collapsed": false,
+      "datasource": null,
+      "gridPos": {
+        "h": 1,
+        "w": 24,
+        "x": 0,
+        "y": 8
+      },
+      "id": 19,
+      "panels": [],
+      "repeat": null,
+      "title": "SLI: Availability",
+      "type": "row"
+    },
+    {
+      "content": "![jupyterhubimage](https://avatars0.githubusercontent.com/u/17927519?s=200&v=4)",
+      "datasource": null,
+      "fieldConfig": {
+        "defaults": {
+          "custom": {}
+        },
+        "overrides": []
+      },
+      "gridPos": {
+        "h": 8,
+        "w": 6,
+        "x": 0,
+        "y": 9
+      },
+      "id": 7,
+      "links": [],
+      "mode": "markdown",
+      "options": {
+        "content": "![jupyterhubimage](https://avatars0.githubusercontent.com/u/17927519?s=200&v=4)",
+        "mode": "markdown"
+      },
+      "pluginVersion": "7.1.0",
+      "title": "",
+      "transparent": true,
+      "type": "text"
+    },
+    {
+      "datasource": "opendatahub",
+      "fieldConfig": {
+        "defaults": {
+          "custom": {},
+          "mappings": [
+            {
+              "from": "",
+              "id": 0,
+              "text": "Up",
+              "to": "",
+              "type": 1,
+              "value": "1"
+            },
+            {
+              "from": "",
+              "id": 1,
+              "text": "Down",
+              "to": "",
+              "type": 1,
+              "value": "0"
+            }
+          ],
+          "thresholds": {
+            "mode": "absolute",
+            "steps": [
+              {
+                "color": "green",
+                "value": null
+              },
+              {
+                "color": "dark-red",
+                "value": 0
+              },
+              {
+                "color": "green",
+                "value": 1
+              }
+            ]
+          }
+        },
+        "overrides": []
+      },
+      "gridPos": {
+        "h": 7,
+        "w": 4,
+        "x": 7,
+        "y": 9
+      },
+      "id": 25,
+      "options": {
+        "colorMode": "background",
+        "graphMode": "none",
+        "justifyMode": "auto",
+        "orientation": "auto",
+        "reduceOptions": {
+          "calcs": [
+            "lastNotNull"
+          ],
+          "fields": "",
+          "values": false
+        },
+        "textMode": "auto"
+      },
+      "pluginVersion": "7.1.1",
+      "targets": [
+        {
+          "expr": "up{service=\"jupyterhub\"}",
+          "interval": "",
+          "legendFormat": "",
+          "refId": "A"
+        },
+        {
+          "refId": "B"
+        }
+      ],
+      "timeFrom": null,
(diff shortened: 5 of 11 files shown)


grafana/grafana/base/kustomization.yaml
```diff
@@ -10,3 +10,5 @@ resources:
 - dashboard.yaml
 - argo-dashboard.yaml
 - grafana.yaml
+- jupyterhub-sli-dashboard.yaml
```
anishasthana, line 13: Could you sort the dashboards in alphabetical order? :-)

hemajv: done :)

grafana/grafana/base/jupyterhub-usage-dashboard.yaml
```diff
@@ -0,0 +1,855 @@
...
+                  ],
+                  "fields": "",
+                  "values": false
+                },
+                "showUnfilled": true
+              },
+              "pluginVersion": "7.1.1",
+              "targets": [
+                {
+                  "expr": "kubelet_volume_stats_used_bytes{namespace=\"opf-jupyterhub\", pod=\"prometheus-k8s-0\"}/kubelet_volume_stats_capacity_bytes{namespace=\"opf-jupyterhub\", pod=\"prometheus-k8s-0\"}",
```
anishasthana, line 147: hmm, `opf-jupyterhub` won't work for us will it?

crobby:
/test all
Seems like some sort of flake with the gcp cluster

grafana/grafana/base/jupyterhub-usage-dashboard.yaml
hemajv: @anishasthana added a `namespace` template variable from which you can select the namespace you want

crobby (commented):
I'm ok with these.  Anyone else have an opinion?

anishasthana:
The JSON itself looks good.

grafana/grafana/base/jupyterhub-usage-dashboard.yaml
```diff
@@ -0,0 +1,881 @@
...
+                  ],
+                  "fields": "",
+                  "values": false
+                },
+                "showUnfilled": true
+              },
+              "pluginVersion": "7.1.1",
+              "targets": [
+                {
+                  "expr": "kubelet_volume_stats_used_bytes{namespace=\"$namespace\", pod=\"prometheus-k8s-0\"}/kubelet_volume_stats_capacity_bytes{namespace=\"$namespace\", pod=\"prometheus-k8s-0\"}",
```
anishasthana, line 147: Is this metric something we have access to on all deployments of ODH? I thought it comes from Kube State Metrics?

anishasthana: Should we just average the value from the two prometheus-k8s pods?

grafana/grafana/base/jupyterhub-usage-dashboard.yaml
```diff
@@ -0,0 +1,881 @@
...
+                "useTags": false
+              },
+              {
+                "allValue": null,
+                "current": {
+                  "selected": false,
+                  "text": "opf-kafka",
+                  "value": "opf-kafka"
+                },
+                "datasource": "$datasource"
```
LaVLaS, line 839: Missing comma at the end of this line

grafana/grafana/base/jupyterhub-usage-dashboard.yaml
```diff
@@ -0,0 +1,881 @@
...
+                "definition": "label_values(container_cpu_usage_seconds_total, namespace)"
```
LaVLaS, line 840: Missing comma at the end of this line

grafana/grafana/base/jupyterhub-sli-dashboard.yaml
```diff
@@ -0,0 +1,1579 @@
...
+                "useTags": false
+              },
+              {
+                "allValue": null,
+                "current": {
+                  "selected": false,
+                  "text": "",
+                  "value": ""
+                },
+                "datasource": "$datasource"
```
LaVLaS, line 1526: Missing comma at the end of line

grafana/grafana/base/jupyterhub-sli-dashboard.yaml
```diff
@@ -0,0 +1,1579 @@
...
+                "definition": "label_values(container_cpu_usage_seconds_total, namespace)"
```
LaVLaS, line 1527: Missing comma at the end of line

anishasthana:
I wonder if we should start federating metrics from cluster monitoring by default. This would allow us to include more useful dashboards/alerts with ODH right out of the box. @hemajv what do you think?

LaVLaS (changes requested):
See comments about missing commas

anishasthana (changes requested):
We will need to federate metrics from kube-state-metrics/ cluster monitoring before this dashboard works.

hemajv:
> I wonder if we should start federating metrics from cluster monitoring by default. This would allow us to include more useful dashboards/alerts with ODH right out of the box. @hemajv what do you think

yes, some of the metrics in this dashboard such as pod, CPU and memory usage are from kube state metrics and these metrics need to be federated

grafana/grafana/base/jupyterhub-usage-dashboard.yaml
4n4nd: they are 2 replicas of the same metric so it doesn't really matter I think

grafana/grafana/base/jupyterhub-usage-dashboard.yaml
hemajv: fixed :+1:

grafana/grafana/base/jupyterhub-usage-dashboard.yaml
hemajv: fixed!

grafana/grafana/base/jupyterhub-sli-dashboard.yaml
hemajv: fixed!

grafana/grafana/base/jupyterhub-sli-dashboard.yaml
hemajv: fixed :+1:

hemajv:
@anishasthana added the cluster metrics service monitor to federate the required metrics in a separate commit: https://github.com/opendatahub-io/odh-manifests/pull/316/commits/504594441c4c1d1a1e5a6092142db64fe7e225c0

prometheus/operator/base/cluster-metrics-servicemonitor.yaml
```diff
@@ -0,0 +1,25 @@
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
+metadata:
+  name: cluster-monitor-federation
+  labels:
```
anishasthana, line 5: We will need the following label on this service monitor
```
        team: opendatahub
```

anishasthana (changes requested):

prometheus/operator/base/cluster-metrics-servicemonitor.yaml
hemajv: added :+1:

prometheus/operator/base/cluster-metrics-servicemonitor.yaml
```diff
@@ -0,0 +1,26 @@
...
+    monitor-component: cluster-resources
+    team: opendatahub
+spec:
+  endpoints:
+    - honorLabels: true
+      params:
+        'match[]':
+          - '{__name__=~".+"}'
```
anishasthana, line 13: I'm sorry I keep nitting on things :sweat_smile: . Do you know specifically what metrics we are using for the dashboards?
Ideally we'd only federate those over from cluster monitoring. This way we could also reduce the scrape intervals

anishasthana (changes requested):

prometheus/operator/base/cluster-metrics-servicemonitor.yaml
hemajv: @anishasthana added the specific metrics that need to be federated :+1:

anishasthana: does 30s seem reasonable?

anishasthana (changes requested):
1. We should split out the dashboard json so that it's easy to re-use for everyone
2. Prometheus should not be able to scrape the cluster monitoring metrics as far as I know, were you able to verify? We may need to give the prometheus SA reader on openshift-monitoring

hemajv:
> 1. We should split out the dashboard json so that it's easy to re-use for everyone
> 2. Prometheus should not be able to scrape the cluster monitoring metrics as far as I know, were you able to verify? We may need to give the prometheus SA reader on openshift-monitoring

1. Yes, agreed. I uploaded the jsons themselves into a separate `grafana/grafana/base/dashboards` folder and updated the GrafanaDashboard resources to include a `url` field which would load the json from the specified path. This way everyone can just reference and have their dashboards point to the JSONs here.
2. I did not get a chance to verify if Prometheus can scrape the cluster monitoring metrics. Will let you know if I do get a chance to do so. However, I do think it might require some additional permissions. @4n4nd would you happen to know more about this?

4n4nd:
> > 1. We should split out the dashboard json so that it's easy to re-use for everyone
> > 2. Prometheus should not be able to scrape the cluster monitoring metrics as far as I know, were you able to verify? We may need to give the prometheus SA reader on openshift-monitoring
> 
> 1. Yes, agreed. I uploaded the jsons themselves into a separate `grafana/grafana/base/dashboards` folder and updated the GrafanaDashboard resources to include a `url` field which would load the json from the specified path. This way everyone can just reference and have their dashboards point to the JSONs here.
> 2. I did not get a chance to verify if Prometheus can scrape the cluster monitoring metrics. Will let you know if I do get a chance to do so. However, I do think it might require some additional permissions. @4n4nd would you happen to know more about this?

@hemajv @anishasthana since the openshift-monitoring Prometheuses are behind a proxy, you need some specific privileges to access the metrics. It's documented [here](https://docs.openshift.com/container-platform/4.5/monitoring/cluster_monitoring/prometheus-alertmanager-and-grafana.html#monitoring-accessing-prometheus-alertmanager-grafana-directly_accessing-prometheus).

anishasthana:
@LaVLaS are we good with giving the cluster-monitoring-view role to the prometheus service account for this PR?

hemajv:
@anishasthana @LaVLaS @nakfour based on our discussion, in the PR I have already stored the JSONs in a separate `dashboards` folder within (the ODH repo) `grafana/grafana/base/dashboards` and referencing to these in the Grafana resource: https://github.com/opendatahub-io/odh-manifests/pull/316/files#diff-01d85d5ce7edad37458697d9a313ba6ee31fe90a49a21d3870b4e125b2d25b7dR8, let me know if there is anything else that needs to be changed

LaVLaS (changes requested):
> @anishasthana @LaVLaS @nakfour based on our discussion, in the PR I have already stored the JSONs in a separate `dashboards` folder within (the ODH repo) `grafana/grafana/base/dashboards` and referencing to these in the Grafana resource: https://github.com/opendatahub-io/odh-manifests/pull/316/files#diff-01d85d5ce7edad37458697d9a313ba6ee31fe90a49a21d3870b4e125b2d25b7dR8, let me know if there is anything else that needs to be changed

Instead of using referencing `odh-manifests` as the remote url for the json, why can we not just embed the json in the `GrafanaDashboard` object directly in the same manner as the [argo dashboard](https://github.com/opendatahub-io/odh-manifests/blob/master/grafana/grafana/base/argo-dashboard.yaml#L9).  We only need to use the url if the json would be stored in a repo other than `odh-manifests`

```yaml
apiVersion: integreatly.org/v1alpha1
kind: GrafanaDashboard
metadata:
  name: odh-jupyterhub-usage
  labels:
    app: grafana
spec:
  json: |
  {
    "annotations": {
      "list": [
        {
          "builtIn": 1,
          "datasource": "-- Grafana --",
          "enable": true,
          "hide": true,
   ...
  }
```

grafana/grafana/base/jupyterhub-sli-dashboard.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: integreatly.org/v1alpha1
+kind: GrafanaDashboard
+metadata:
+  name: odh-jupyterhub-sli
+  labels:
+    app: grafana
+spec:
+  url: https://raw.githubusercontent.com/odh-manifests/grafana/grafana/base/dashboards/jupyterhub-sli-slo.json
```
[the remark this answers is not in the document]
LaVLaS: I agree on linking to a static sha or tag when loading a json object from a url.  In this instance, I don't think we need the `url` since this json would be hosted in `odh-manifests`.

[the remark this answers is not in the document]
hemajv: So the idea of having the jsons for ODH related component dashboards stored in a separate folder in `odh-manifests` was that if these dashboards are "basic/generic" enough, then we (operate-first) can point to these jsons as a remote URL in our grafana resource and if we need to customize the dashboard further for our needs, we can take the json and store the modified jsons in our SRE repo.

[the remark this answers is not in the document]
4n4nd: I think the reason for keeping the dashboard jsons separate was because some installations like IDH were not using the Grafana Operator and they were planning to use them directly, @anishasthana and @accorvin might know better.

I think we can put the jsons in a configmap and reference them in the `GrafanaDashboard` object ([doc](https://github.com/integr8ly/grafana-operator/blob/master/documentation/dashboards.md#config-map-references)), this way IDH should be able to directly reference the dashboard.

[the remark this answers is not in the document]
anishasthana: ++ for keeping the jsons separate. That makes life a lot easier for us downstream.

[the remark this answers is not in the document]
LaVLaS: After an offline conversation with @hemajv explaining the reasoning for separating the json from the `GrafanaDashboard`, I understand now. 

We need to use a tagged reference to this dashboard url and, as far as I know, we can't do that within this PR since it is adding the files we need to "tag".  From @4n4nd's comment, we should use a `ConfigMap` reference instead of the url so that operate-first, downstream, ... can override it easily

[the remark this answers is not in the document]
hemajv: shall I go ahead and update the PR to have the dashboard jsons embedded in the `ConfigMap` then rather than saving the individual json files? @anishasthana wdyt?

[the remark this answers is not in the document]
anishasthana: Hmmm, configmap generator + JSONS are easier to work with as developers. indented jsons in yaml are pretty miserable -- any cahnce we could do that instead, @LaVLaS ? If not, I'm fine with embedding them directly in the CM.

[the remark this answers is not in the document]
LaVLaS: I have no issue with the ConfigMapGenerator referencing pure json

LaVLaS:
The `GrafanaDashboards` are created successfully but they appear unable to retrieve metrics for any JupyterHub notebooks.

hemajv:
> The `GrafanaDashboards` are created successfully but they appear unable to retrieve metrics for any JupyterHub notebooks.

Update the PR with the necessary fixes. The prometheus token needed to be added to the service monitor in order for the metrics to be federated. I have also added the `cluster-monitoring-view` role/rolebindings required. 
cc @LaVLaS @anishasthana

prometheus/operator/base/cluster-metrics-servicemonitor.yaml
```diff
@@ -0,0 +1,41 @@
...
+  name: cluster-monitor-federation
+  labels:
+    monitor-component: cluster-resources
+    team: opendatahub
+spec:
+  endpoints:
+    - bearerTokenSecret:
+        key: token
+        name: odh.prometheus-k8s
+    - honorLabels: true
```
4n4nd, line 13: There is a stray `-` here, there should only be one endpoint object

hemajv: thanks @4n4nd! :pray:

4n4nd (approved):
/lgtm

LaVLaS (approved):
/lgtm

Dashboards are pulling JH server and user info for all spawned pods

anishasthana (approved):
/lgtm

PR #329, opened 2021-02-23
Remove tls-acme annotation from JuptyerHub route

accorvin:
This annotation configures Kubernetes to automatically provision TLS
certificates for the JupyterHub route from Let's Encrypt. In my opinion,
we should rely on OpenShift-provisioned certs by default and let users
override this as they need, whether that means ACME or something else.


crobby (approved):
lgtm

crobby:
/test odh-manifests-e2e
gcp flake

Issue #310, opened 2021-02-10
Upgrade from v0.9 to v1.0.0: JupyterHub DeploymentConfig is missing ImageChange trigger


tumido:
**Describe the bug**
JupyterHub DeploymentConfig is missing ImageChange trigger, therefore a migration updating the `jupyterhub-img` `ImageStream` is not propagating via a new rollout.

**Steps To Reproduce**
1. Have a `v0.9` deployment available
2. Update the repository to `v1.0.0`
3. This applies https://github.com/opendatahub-io/odh-manifests/pull/245 PR
4. IImageStream `jupyterhub-img` is updated
5. DeploymentConfig `jupyterhub` is not rolling out

**Expected behavior**
A rollout is triggered

**Workaround (if any)**
Trigger a manual rollout.

**Open Data Hub Version**
Migrating from `v0.9` to `v1.0.0`

**OpenShift Version**
Version #: 4.5
Provider (Baremetal, OpenStack, RHV, AWS, OKD, CodeReady Containers, ...): OpenStack

**Additional context**
https://github.com/operate-first/apps/issues/231#issuecomment-776758568

PR #334, opened 2021-03-01
Added ImageChange trigger to the JupyterHub deployment config

mroman-redhat:
The trigger is based on an older implementation: https://gitlab.com/opendatahub/jupyterhub-ansible/-/blob/master/templates/jupyterhub-deployer/jupyterhub-dc.json.j2#L15
It should be changed later so it does not use :latest image.

Closes: #310


vpavlin:
I believe we removed this in the past to avoid issues with endless redeployments. We need to carefully verify this before we add it back

jupyterhub/jupyterhub/base/jupyterhub-dc.yaml
```diff
@@ -93,3 +93,12 @@ spec:
         name: config
   triggers:
   - type: ConfigChange
+  - type: ImageChange
+    imageChangeParams: 
+      automatic: true
+      containerNames:
+        -wait-for-database
```
LaVLaS, line 100: Currently this is being parsed as a string, this needs a space after the `-` to make it a list

LaVLaS (changes requested):

LaVLaS (approved):

PR #342, opened 2021-03-08
Add JH groups configuration CM

vpavlin:
This PR adds a configmap to set names of groups for JH users and admins

1. By default, no groups are used and anyone with OpenShift login have access to JH
2. User can provide a custom CM with groups listed and set the name of the CM as `parameter` - `jupyterhub_groups_config
`

The secret is then managed outside of the operator.

### DIFF
--- a/jupyterhub/README.md
+++ b/jupyterhub/README.md
@@ -21,6 +21,10 @@ HTTP endpoint exposed by your S3 object storage solution which will be made avai
 
 Name of the storage class to be used for PVCs created by JupyterHub component. This requires `storage-class` **overlay** to be enabled as well to work.
 
+#### jupyterhub_groups_config
+
+A ConfigMap containing comma separated lists of groups which would be used as Admin and User groups for JupyterHub. The default ConfgiMap can be found [here](jupyterhub/base/jupyterhub-groups-configmap.yaml).
+
 ##### Examples
 
 ```
--- a/jupyterhub/jupyterhub/base/jupyterhub-dc.yaml
+++ b/jupyterhub/jupyterhub/base/jupyterhub-dc.yaml
@@ -63,6 +63,16 @@ spec:
             configMapKeyRef:
               name: jupyterhub-cfg
               key: singleuser_pvc_size
+        - name: JUPYTERHUB_ADMIN_GROUPS
+          valueFrom:
+            configMapKeyRef:
+              name: $(jupyterhub_groups_config)
+              key: admin_groups
+        - name: JUPYTERHUB_ALLOWED_GROUPS
+          valueFrom:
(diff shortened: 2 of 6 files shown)


accorvin:
cc @wseaton for awareness

crobby:
> This PR adds a configmap to set names of groups for JH users and admins
> 
> 1. By default, no groups are used and anyone with OpenShift login have access to JH
> 2. User can provide a custom Secret with groups listed and set the name of the secret as `parameter` - `jupyterhub_groups_config `
> 
> The secret is then managed outside of the operator.



> This PR adds a configmap to set names of groups for JH users and admins
> 
> 1. By default, no groups are used and anyone with OpenShift login have access to JH
> 2. User can provide a custom Secret with groups listed and set the name of the secret as `parameter` - `jupyterhub_groups_config `
> 
> The secret is then managed outside of the operator.

So...the defaults are stored in a configMap, but if the user wants something custom, that gets stored in a Secret?

accorvin:
also cc @durandom

vpavlin:
> So...the defaults are stored in a configMap, but if the user wants something custom, that gets stored in a Secret?

Ups..typo..fixed the description:) Sorry:)

crobby (commented):
I think this is probably good.  I'd like another person to also have a look though.

tumido:
@vpavlin sure! I can take a look and @4n4nd can probably as well. :slightly_smiling_face:  I need to test it out, thank you!

tumido:
Quick question:
If the values are left empty what's does it default to?

Am I right to assume:
- `admin_groups` default to none (or cluster admin?)
- `allowed_groups` default to `system:authenticated`?

wseaton:
@tumido I helped work on the upstream PR so I can help answer these:

>     admin_groups default to none (or cluster admin?)

In my testing this defaults to no admins

>     allowed_groups default to system:authenticated?

Yep! Any user who can authenticate with the cluster can log into the hub.

accorvin:
> I remember there was a group called something like opendatahub-admins that was created by the odh operator, would it make sense to set that as the default value for admin_groups?

I'm pretty sure that was a group that we manually created in our internal deployment, not something that the ODH operator created. I may be wrong though.

vpavlin:
Yup, we do not have any default groups. We have a user `admin` be an admin by default, but otherwise we leave this up to the user

vpavlin:
@crobby Can you approve this? Seems we have bunch of lgtms

tumido (commented):
/lgtm

vpavlin:
@crobby The tests failed because of Thrift server (IIUC). If they fail again, can we merge manually? We need this in the release tomorrow

PR #344, opened 2021-03-09
WIP: Add oauth proxy to ODH dashboard

accorvin:
This adds a basic OAuth Proxy sidecar container to the ODH Dashboard
pod. By default this will allow any authenticated user to access the
service. I expect downstream deployments to limit access by way of the
--openshift-sar argument. I additionally expect downstream deployments
to customize the value of the --cookie-secret argument.


odh-dashboard/base/deployment.yaml
```diff
@@ -14,6 +14,31 @@ spec:
...
       containers:
+      - name: oauth-proxy
+        args:
+        - --https-address=:8443
+        - --provider=openshift
+        - --openshift-service-account=odh-dashboard
+        - --upstream=http://localhost:8080
+        - --tls-cert=/etc/tls/private/tls.crt
+        - --tls-key=/etc/tls/private/tls.key
+        - --cookie-secret=SECRET
```
vpavlin, line 25: Do we need to parameterize te SECRET?

accorvin: likely, yes. I could use a pointer to an example for how to do that.

Downstream deployments will use a slightly different method to generate a random value of this (see https://github.com/red-hat-data-services/odh-deployer/blob/main/deploy.sh#L34)

vpavlin:
Should we have access controlled by groups here as well?

I think we could use https://github.com/openshift/oauth-proxy#require-specific-permissions-to-login-via-oauth-with---openshift-sarjson

And define: https://docs.openshift.com/container-platform/4.5/rest_api/authorization_apis/subjectaccessreview-authorization-openshift-io-v1.html

I am not sure how to define the list of groups though as they are in JH ConfigMap and here we'd probably have to do it via parameters

odh-dashboard/base/deployment.yaml
vpavlin: Hmm...we do not have a way to generate a random value at the moment.

Here is an example of a parameter:
- https://github.com/opendatahub-io/odh-manifests/blob/master/jupyterhub/jupyterhub/base/jupyterhub-singleuser-profiles-configmap.yaml#L11
- https://github.com/opendatahub-io/odh-manifests/blob/master/jupyterhub/jupyterhub/base/params.env#L2
- For non-standard fields you need to add kind and path here: https://github.com/opendatahub-io/odh-manifests/blob/master/jupyterhub/jupyterhub/base/params.yaml
- https://github.com/opendatahub-io/odh-manifests/blob/master/jupyterhub/jupyterhub/base/kustomization.yaml#L43-L49 (the CM `parameters` is generated based on `params.env`)

crobby:
I think the tests will likely fail unless they are updated a bit:  https://github.com/opendatahub-io/odh-manifests/blob/master/tests/basictests/dashboard.sh#L29

odh-dashboard/overlays/authentication/deployment.yaml
```diff
@@ -0,0 +1,34 @@
+- op: add
+  path: /spec/template/spec/containers/-
+  value:
+    - name: oauth-proxy
```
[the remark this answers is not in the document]
accorvin: Yep, I'm still working on testing this. I'll let you know once I have it working again and this is ready for reviews once more.

accorvin:
Closing until I get the tests working, will reopen.

accorvin:
I've reopened this branch for PR in #350 

All comments here should be resolved there (and I'm hoping the tests will pass)

PR #355, opened 2021-03-22
Add Trino kustomize manifests

rimolive:


LaVLaS (changes requested):
@rimolive This will need some basictests that we can include in CI runs to verify it is working as intended.

nakfour:
@rimolive can you please add a description for what this PR is adding and how to test it

anishasthana:
https://github.com/opendatahub-io/odh-manifests/pull/373 replaces this

PR #361, opened 2021-03-24
Update JSP to 0.2.2, add secrets to the role

vpavlin:

PR #371, opened 2021-03-29
Modified jhtest to accept new version of JH Spawner

mroman-redhat:
This change depends on this patch to JSP: https://github.com/opendatahub-io/jupyterhub-singleuser-profiles/pull/106

Note: Had to add an additional if statement on line 183 to make the tests work. For some reason the Jupyterhub 'Sign in with openshift' button was not being clicked. I do not know whether this will break anything @vpavlin .

### DIFF
--- a/tests/basictests/jupyterhub/jhtest.py
+++ b/tests/basictests/jupyterhub/jhtest.py
@@ -76,7 +76,7 @@ class JHStress():
 
     def go_to_admin(self):
         if not self.check_exists_by_xpath("//a[@href='/hub/admin']"):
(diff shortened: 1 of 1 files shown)


crobby:
Seems like the error might be "real":  no such element: Unable to locate element: {"method":"css selector","selector":"[id="KeyForm-"]"}

mroman-redhat:
@crobby I needs to be pushed on top of jsp 0.2.2, it had the incorrect version of the spawner in the screenshot. We are testing it in #361

vpavlin:
This can be closed - I used @mroman-redhat's commin in the 0.2.2 PR

PR #373, opened 2021-03-29
Add Trino kustomize manifests

maulikjs:
This PR adds the deployment artifacts for Trino Controller and worker nodes. It also introduces some config options for the kfctl and adds a test for the deployment.


crobby:
/test all
Hoping it was a flake

crobby:
Hmm, same 3 errors as before (jupyterhub, prometheus, and trino)

nakfour:
@maulikjs please describe what this feature provides and how to test this, and does it need test use cases.

trino/base/hive-metastore-dc.yaml
```diff
@@ -0,0 +1,148 @@
...
+                key: database-password
+                name: trino-db-secret
+          - name: DATABASE_HOST
+            value: trino-db
+          - name: DATABASE_PORT
+            value: 5432
+          - name: DATABASE_NAME
+            value: trino
+          - name: DATABASE_SSLMODE
+            value: accept
```
rimolive, line 66: Checking PostgreSQL documentation I can't find this value. Valid options are: `disable`, `allow`, `prefer`, `require`, `verify-ca` and `verify-full`. Source: https://www.postgresql.org/docs/9.1/libpq-ssl.html

trino/base/hive-metastore-dc.yaml
```diff
@@ -0,0 +1,148 @@
...
+                name: trino-db-secret
+          - name: DATABASE_PASSWORD
+            valueFrom:
+              secretKeyRef:
+                key: database-password
+                name: trino-db-secret
+          - name: DATABASE_HOST
+            value: trino-db
+          - name: DATABASE_PORT
+            value: 5432
```
rimolive, line 62: Env var values must be enclosed with double quotes, otherwise kfctl will fail.

trino/base/hive-metastore-dc.yaml
```diff
@@ -0,0 +1,148 @@
...
+          - name: DATABASE_NAME
+            value: trino
```
rimolive, line 64: Env var values must be enclosed with double quotes, otherwise kfctl will fail.

rimolive (changes requested):

rimolive:
@nakfour Trino is a SQL Analytics database with multiple data sources support. We expect this to be the replacement of Thrift Server in the Data Catalog in the future. The basic test only checks for the objects and if the pods are in Running state, but I'll provide the steps to test this.

trino/base/hive-metastore-dc.yaml
accorvin: Typo. Fixed.

rimolive:
Instructions to test:

- Follow CLI installation docs in the URL: https://trino.io/docs/current/installation/cli.html#installation
- Run the CLI command to connect to Trino: `trino --server $(oc get route trino-route -o jsonpath='{$.spec.host}'):80 --catalog hive --schema default`
- When the trino prompt appears, run `select 1;` SQL statement
- The result should be like this:
```
 _col0 
-------
     1 
(1 row)

Query 20210407_171244_00004_pbhx8, FINISHED, 1 node
Splits: 1 total, 1 done (100.00%)
0.52 [0 rows, 0B] [0 rows/s, 0B/s]
```

trino/base/hive-metastore-dc.yaml
accorvin: Fixed.

trino/base/hive-metastore-dc.yaml
accorvin: Fixed.

accorvin:
Something is still causing the hive metastore to restart regularly. I haven't figured out yet why that's happening. This appears to be working otherwise.

trino/base/hive-metastore-dc.yaml
```diff
@@ -0,0 +1,152 @@
...
+            value: "INFO"
+          - name: S3_ENDPOINT
+            value: $(s3_endpoint_url)
+          - name: S3_ENDPOINT_URL_PREFIX
+            value: $(s3_endpoint_url_prefix)
+          - name: AWS_ACCESS_KEY_ID
+            valueFrom:
+              secretKeyRef:
+                key: AWS_ACCESS_KEY_ID
+                name: aws-secret
```
rimolive, line 40: Hard-coded secret name. It should be set to `$(s3_credentials_secret)`

trino/base/hive-metastore-dc.yaml
```diff
@@ -0,0 +1,152 @@
...
+          - name: AWS_SECRET_ACCESS_KEY
+            valueFrom:
+              secretKeyRef:
+                key: AWS_SECRET_ACCESS_KEY
+                name: aws-secret
```
rimolive, line 45: Hard-coded secret name. It should be set to `$(s3_credentials_secret)`

trino/base/hive-metastore-dc.yaml
```diff
@@ -0,0 +1,152 @@
...
+          - name: S3_BUCKET_NAME
+            valueFrom:
+              secretKeyRef:
+                key: BUCKET
+                name: aws-secret
```
rimolive, line 50: Hard-coded secret name. It should be set to `$(s3_credentials_secret)`

rimolive (changes requested):

rimolive:
errrr.... /lgtm again?

trino/base/trino-db-dc.yaml
```diff
@@ -0,0 +1,66 @@
...
+            - name: POSTGRESQL_PASSWORD
+              valueFrom:
+                secretKeyRef:
+                  name: $(trino_db_secret)
+                  key: database-password
+            - name: POSTGRESQL_PASSWORD
+              valueFrom:
+                secretKeyRef:
+                  name: $(trino_db_secret)
+                  key: database-name
```
LaVLaS, line 50: The key `database-name` is not present in the secret `trino-db-secret`

LaVLaS (changes requested):

trino/base/trino-db-dc.yaml
maulikjs: fixed

maulikjs: fixed

LaVLaS (approved):

LaVLaS:
> Something is still causing the hive metastore to restart regularly. I haven't figured out yet why that's happening. This appears to be working otherwise.

The `hive-metastore-0` pod appears to be stuck in a crashloop because `XXX_S3ENDPOINT_URL_PREFIX_XXX` isn't being replaced correctly in `hive-metastore.xml`
```
21/04/14 00:48:41 [main]: ERROR s3a.S3AFileSystem: Incorrect endpoint: java.net.URISyntaxException: Illegal character in scheme name at index 3: XXX_S3ENDPOINT_URL_PREFIX_XXXhttps:///
21/04/14 00:48:41 [main]: ERROR metastore.RetryingHMSHandler: java.lang.IllegalArgumentException: Incorrect endpoint: java.net.URISyntaxException: Illegal character in scheme name at index 3: XXX_S3ENDPOINT_URL_PREFIX_XXXhttps:///
	at org.apache.hadoop.fs.s3a.DefaultS3ClientFactory.createAmazonS3Client(DefaultS3ClientFactory.java:206)
	at org.apache.hadoop.fs.s3a.DefaultS3ClientFactory.createS3Client(DefaultS3ClientFactory.java:55)
	at org.apache.hadoop.fs.s3a.S3AFileSystem.initialize(S3AFileSystem.java:256)
	at org.apache.hadoop.fs.FileSystem.createFileSystem(FileSystem.java:3354)
	at org.apache.hadoop.fs.FileSystem.access$200(FileSystem.java:124)
	at org.apache.hadoop.fs.FileSystem$Cache.getInternal(FileSystem.java:3403)
...
```

rimolive (approved):

trino/base/hive-metastore-entrypoint.yaml
```diff
@@ -0,0 +1,120 @@
...
+    fi
+    if [ -f /hadoop-config/hdfs-site.xml ]; then
+      ln -s -f /hadoop-config/hdfs-site.xml /etc/hadoop/hdfs-site.xml
+    else
+      echo "hdfs-site.xml doesnt exist, skipping symlink"
+    fi
+
+    # insert S3 bucket URI from env vars
+    cat /hive-config/hive-site.xml | sed \
+      -e "s#XXX_S3LOCATION_XXX#${S3_ENDPOINT_URL_PREFIX}#" \
```
LaVLaS, line 70: This appears to be a bug since you are replacing `XXX_S3LOCATION_XXX` twice with 2 different values.
I think you wanted to set this to 
```
-e "s#XXX_S3ENDPOINT_URL_PREFIX_XXX#${S3_ENDPOINT_URL_PREFIX}#" \
```

LaVLaS (changes requested):

trino/base/hive-metastore-entrypoint.yaml
maulikjs: yes you are correct, I messed it up while staging changes 🤦

trino/base/hive-config.yaml
```diff
@@ -0,0 +1,189 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: hive-config
+data:
+  hive-exec-log4j2.properties: 'status = INFO
+
```
tumido, line 7: You can use Multiline YAML strings here and you don't have to leave the blank lines in between:

```yaml
a: |
  each
  of
  these
  words
  are
  on
  separate
  lines
```

https://yaml-multiline.info/

trino/base/hive-config.yaml
```diff
@@ -0,0 +1,189 @@
...
+    # root logger
+
+    rootLogger.level = ${sys:hive.log.level}
+
+    rootLogger.appenderRefs = root
+
+    rootLogger.appenderRef.root.ref = ${sys:hive.root.logger}
+
+    '
+  hive-site.xml: "<configuration>\n  <property>\n    <name>hive.server2.enable.doAs</name>\n\
```
tumido, line 156: multiline YAML strings should be used here as well

trino/base/hive-jmx-config.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  labels:
+    app: hive
+  name: hive-jmx-config
+data:
+  config.yml: "---\nlowercaseOutputName: true\nlowercaseOutputLabelNames: true\n\
```
tumido, line 8: multiline YAML strings should be used here as well

trino/base/trino-db-dc.yaml
```diff
@@ -0,0 +1,66 @@
...
+          resources:
+            limits:
+              cpu: 1
+              memory: "2Gi"
+            requests:
+              cpu: 1
```
tumido, line 56: This CPU request seems a bit high for a DB pod. No need for DB to reserve whole cpu.

trino/base/params.env
```diff
@@ -0,0 +1,14 @@
+s3_endpoint_url=changeme
+s3_endpoint_url_prefix=https://
+s3_credentials_secret=aws-secret
+trino_environment=trino
+trino_db_secret=trino-db-secret
+hive_metastore_cpu_request=1
```
tumido, line 6: I don't think the CPU requests needs to be this high. Keep in mind that request is reserved even if the pod is fully idle.

trino/base/params.env
```diff
@@ -0,0 +1,14 @@
...
+hive_metastore_cpu_limit=1
+hive_metastore_memory_request=1Gi
+hive_metastore_memory_limit=1Gi
+trino_cpu_request=1
```
tumido, line 10: Same as above.

trino/base/hive-metastore-entrypoint.yaml
```diff
@@ -0,0 +1,120 @@
...
+
+    # insert S3 bucket URI from env vars
+    cat /hive-config/hive-site.xml | sed \
+      -e "s#XXX_S3ENDPOINT_URL_PREFIX_XXX#${S3_ENDPOINT_URL_PREFIX}#" \
+      -e "s#XXX_S3LOCATION_XXX#${S3_ENDPOINT}#" \
+      -e "s#XXX_S3ENDPOINT_XXX#${S3_BUCKET_NAME}/${S3_DATA_DIR}#" \
+      -e "s#XXX_DATABASE_USER_XXX#${DATABASE_USER}#" \
+      -e "s#XXX_DATABASE_PASSWORD_XXX#${DATABASE_PASSWORD}#" \
+      -e "s#XXX_DATABASE_CONNECT_URL_XXX#postgresql://${DATABASE_HOST}:${DATABASE_PORT}/${DATABASE_NAME}?sslmode=${DATABASE_SSLMODE}#" \
+    > $HIVE_HOME/conf/hive-site.xml
```
tumido, line 76: Why don't you replace the values directly from kustomize vars in `hive-site.xml`?

If I understand this properly, you first load the values from a configmap/secret into a kustomize var, then you mount it to deployment's environment variables as a plaintext, literal value and then `sed` it into a configmap here? Wouldn't it be better to template the `hive-site.xml` directly via kustomize vars?

Or, if you really want to `sed` it during the entry script, mount the configmap/secret value directly to the environment?
https://github.com/opendatahub-io/odh-manifests/pull/373/files#diff-1781d34c25957ba53765bcf4e45efd79b6463410d5797be42c4cccb73a01a16eR32-R35

trino/base/params.env
rimolive: I tend to disagree with that because this is just a default value for a parameter so user can set this value to what is better to their environments, and I can tell you that users might need to increase this value since this is a Java application and we don't have any evidences that Java behaves well with less than 1 CPU. It's too risky say that request can be 500m or less because of the JVM implementation.

trino/base/params.env
rimolive: Same as above.

trino/base/params.env
tumido: Interesting. Do you have any links to explanation or deepdives why a full CPU core has to be fully reserved for a Java pod?

I surely don't have as much experience with this as you do, but it doesn't seem right to me at all. Thriftserver in ODH has CPU request of 400m and is running just fine:
https://github.com/opendatahub-io/odh-manifests/blob/9e0f7842558b41098d89653c0b6f74a30cb8736d/thriftserver/thriftserver/base/thriftserver.yaml#L56

In idle it rests at 0.001 CPU for us, and there's no reason for it to reserve full CPU in that time either...

rimolive: There are many things we can discuss and I don't think we should start this discussion here, but let me give you an example: Let's say we use Serial Collector for the GC algorithm (and this is the (recommendation)[https://developers.redhat.com/blog/2017/04/04/openjdk-and-containers/] for 1 CPU), this collector thread will work concurrently with the application threads and thus running what we call a stop-the-world collection to free up the memory. We don't have any benchmarks with Hive Metastore working with 1GB of memory, but considering that 1GB is the total amount of memory reserved for the pod and JVM has other memory areas to manage, the Garbage Collection task can be frequent in this scenario and thus triggering a lot of GC tasks. Having that said, I'm not saying one or another configuration is wrong, but in order to find the best value for resource requests, we need a separate benchmark test to evaluate other variables.

tumido: Hm, are you disputing memory? I'm discussing **CPU reservation** here, not memory.

I don't see recommended CPU requests in the article anywhere. I have no problem with the memory request, memory is usually very cheap on clusters... CPU is what I think is too high for a reservation. I can't find any proof that lower CPU requests may cause Java containers troubles, since the cluster allows seamless scaling up to the CPU limit if needed.

tumido (commented):
/lgtm

trino/base/trino-db-dc.yaml
maulikjs: fixed!

LaVLaS (approved):

maulikjs:
```
Running operator-tests/odh-manifests/basictests/jupyterhub.sh:24: executing 'oc get pods -l deploymentconfig=jupyterhub --field-selector='status.phase=Running' -o jsonpath='{$.items[*].metadata.name}'' expecting any result and text 'jupyterhub'; re-trying every 10s until completion or 1200.000s...
{"component":"entrypoint","file":"prow/entrypoint/run.go:165","func":"k8s.io/test-infra/prow/entrypoint.Options.ExecuteProcess","level":"error","msg":"Process did not finish before 2h0m0s timeout","severity":"error","time":"2021-04-15T02:22:29Z"}
{"component":"entrypoint","file":"prow/entrypoint/run.go:255","func":"k8s.io/test-infra/prow/entrypoint.gracefullyTerminate","level":"error","msg":"Process did not exit before 15s grace period","severity":"error","time":"2021-04-15T02:22:44Z"}
{"component":"entrypoint","error":"process timed out","file":"prow/entrypoint/run.go:80","func":"k8s.io/test-infra/prow/entrypoint.Options.Run","level":"error","msg":"Error executing test process","severity":"error","time":"2021-04-15T02:22:44Z"}
error: failed to execute wrapped command: exit status 127
```

maulikjs:
Looks like the e2e test is flaking out waiting for JH to come up.

LaVLaS:
The last 6 runs failed due to:
1 - CI infra issues
5 - ODH operator was installed, kfdef deployed but multiple components failed basictests due to waiting the maximum amount of pods to reach a `Running` state. 

The 5 CI failures could be due to cluster resource issues but we won't know unless we can debug the CI run live.  I think the `trino` deployment in the CI environment is using more resources than are available and, if so, we need to add the parameter to reduce the `trino_memory_request < 4Gi`.

anishasthana (approved):
/lgtm

PR #383, opened 2021-04-07
add connects-to annotation from jupyterhub to jupyterhub-db

strangiato:
This update adds the connects-to annotation to the JupyterHub pod, which creates a connection in the Topology UI to show a visual connection between the `jupyuterhub` and `jupyterhub-db` pods.

This update does not change any functionality and is only a visual indicator that the pods communicate with each other.


LaVLaS (approved):
/lgtm

PR #368, opened 2021-03-26
include new ubi8 cuda-11.0.3 supporting python3.8 notebooks

harshad16:
include new ubi8 cuda-11.0.3 supporting python3.8 notebooks
Signed-off-by: Harshad Reddy Nalla <hnalla@redhat.com>

PR #384, opened 2021-04-08
configure buildconfig ref based on tags for stability

harshad16:
configure buildconfig ref based on tags for stability

maybe related-to: https://github.com/opendatahub-io/odh-manifests/pull/368#issuecomment-814847584


crobby:
I think this is good and makes sense.  I'd prefer a second opinion though just in case I'm missing something.

LaVLaS:
> I think this is good and makes sense. I'd prefer a second opinion though just in case I'm missing something.

+1. This will ensure that local builds will be have the same features enabled as the pre-built notebooks.

crobby (approved):

crobby (approved):

crobby (approved):

PR #390, opened 2021-04-13
annotation set with image name and python dependency details

harshad16:
annotation set with image name and python dependency details

### DIFF
--- a/jupyterhub/notebook-images/overlays/additional/generic-data-science-notebook-imagestream.yaml
+++ b/jupyterhub/notebook-images/overlays/additional/generic-data-science-notebook-imagestream.yaml
@@ -5,7 +5,7 @@ metadata:
     opendatahub.io/notebook-image: "true"
   annotations:
     opendatahub.io/notebook-image-url: "https://github.com/thoth-station/s2i-generic-data-science-notebook"
-    opendatahub.io/notebook-image-name: "Generic Data Scient Notebook Image"
+    opendatahub.io/notebook-image-name: "Standard Data Science"
     opendatahub.io/notebook-image-desc: "Jupyter notebook image with a set of data science libraries that advanced AI/ML notebooks will use as a base image to provide a standard for libraries avialable in all notebooks"
   name: s2i-generic-data-science-notebook
 spec:
@@ -13,6 +13,8 @@ spec:
     local: true
   tags:
   - annotations:
+      opendatahub.io/notebook-software: '[{"name":"Python","version":"v3.8.3"}]'
+      opendatahub.io/notebook-python-dependencies: '[{"name":"Boto3","version":"1.17.11"},{"name":"Kafka-Python","version":"2.0.2"},{"name":"Matplotlib","version":"3.1.3"},{"name":"Numpy","version":"1.20.2"},{"name":"Pandas","version":"1.2.3"},{"name":"Scipy","version":"1.6.2"}]'
       openshift.io/imported-from: quay.io/thoth-station/s2i-generic-data-science-notebook
     from:
       kind: DockerImage


jeff-phillips-18:
The UX design also shows a short description for each (lighter grey text):
image

@kywalker-rh Can you provide insight into where these come from or what they should be?

jupyterhub/notebook-images/overlays/additional/generic-data-science-notebook-imagestream.yaml
```diff
@@ -5,7 +5,8 @@ metadata:
     opendatahub.io/notebook-image: "true"
   annotations:
     opendatahub.io/notebook-image-url: "https://github.com/thoth-station/s2i-generic-data-science-notebook"
-    opendatahub.io/notebook-image-name: "Generic Data Scient Notebook Image"
+    opendatahub.io/notebook-image-name: "Standard Data Science"
+    opendatahub.io/notebook-python-dependencies: [{"name":"Boto3","version":"1.17.11"},{"name":"Kafka-Python","version":"2.0.2"},{"name":"Matplotlib","version":"3.1.3"},{"name":"Numpy","version":"1.20.2"},{"name":"Pandas","version":"1.2.3"},{"name":"Scipy","version":"1.6.2"}]
```
vpavlin, line 9: Looking into this, I think this annotation needs to be on the `tags` level - the reason is that if we need to provide multiple versions of DS notebook, we need to make sure we can provide info for each of them - and they'd probably live in different tags - WDYT?

`spec.tags.annotations`:

* `opendatahub.io/notebook-python-dependencies`
* `opendatahub.io/notebook-tag-name`

I can imagine others like `cuda-compatibility` `verified-gpus` etc..

harshad16: i agree, we can have different versions.

kywalker-rh: @harshad16 Do you define somewhere what software and versions are in the notebook image as well as the python dependancies? I think that is what @jeff-philips-18 is looking for in this case.

harshad16: i have updated it now

harshad16:
@kywalker-rh  @jeff-phillips-18  PTAL 
does this format works? list of dicts with name and version.

jeff-phillips-18:
> @kywalker-rh @jeff-phillips-18 PTAL
> does this format works? list of dicts with name and version.

I'm not sure all entries are name/version pairs. One entry in the mocks shows `Provided by Anaconda`
@kywalker-rh thoughts?

harshad16:
> I'm not sure all entries are name/version pairs. One entry in the mocks shows `Provided by Anaconda`

that would fit as well, [{"name":`Provided by Anaconda`, "version":""}]
in this way, on UI side you can directly use the list of object and set the value by e.name e.version

jupyterhub/notebook-images/overlays/additional/generic-data-science-notebook-imagestream.yaml
```diff
@@ -5,14 +5,16 @@ metadata:
...
-    opendatahub.io/notebook-image-name: "Generic Data Scient Notebook Image"
+    opendatahub.io/notebook-image-name: "Standard Data Science"
     opendatahub.io/notebook-image-desc: "Jupyter notebook image with a set of data science libraries that advanced AI/ML notebooks will use as a base image to provide a standard for libraries avialable in all notebooks"
   name: s2i-generic-data-science-notebook
 spec:
   lookupPolicy:
     local: true
   tags:
   - annotations:
+      opendatahub.io/notebook-software: [{"name":"Python","version":"v3.8.3"}]
```
LaVLaS, line 16: I'm getting the error below when applying these annotations.  It looks like annotations values values can only be strings so these need to be single quoted in order to apply correctly

LaVLaS (changes requested):

jeff-phillips-18:
> > I'm not sure all entries are name/version pairs. One entry in the mocks shows `Provided by Anaconda`
> 
> that would fit as well, [{"name":`Provided by Anaconda`, "version":""}]
> in this way, on UI side you can directly use the list of object and set the value by e.name e.version

OK, that's fine then.

LaVLaS (approved):

harshad16:
> OK, that's fine then.

i was just suggesting, please check on the UI side if this fits. we can modify this as per UI requirements

jeff-phillips-18:
> > OK, that's fine then.
> 
> i was just suggesting, please check on the UI side if this fits. we can modify this as per UI requirements

I believe this will work for the UI.

PR #411, opened 2021-04-22
Add latest tag for s2i-lab-elyra JH image

4n4nd:
This will add a `latest` tag for the `s2i-lab-elyra` image which will update regularly.

cc @ptitzler


LaVLaS:
The `latest` tag should not be included in an ODH manifest.  

When https://github.com/opendatahub-io/opendatahub-operator/issues/123 is implemented, you can add the label `opendatahub.io/modified: true` to the elyra imagestream in your local deployment to prevent the operator from resetting this imagestreamtag

PR #439, opened 2021-06-18
Change Trino Route to a secured one

rimolive:
This change is required for users who want to configure authentication in Trino (it is unsecured by default).


accorvin (commented):
/lgtm

maulikjs (approved):

PR #440, opened 2021-06-18
Add builds and buildconfigs read permissions to odh-dashboard role

jeff-phillips-18:
This change is required because the dashboard is watching build statuses to notify the user of builds in progress, failure, and completion.


LaVLaS (approved):
/lgtm

PR #446, opened 2021-06-22
Add a general cluster role to JH

mroman-redhat:
Added rights necessary for: https://github.com/opendatahub-io/jupyterhub-singleuser-profiles/pull/157
Replaced node reader rights with a general jupyterhub-cluster role


LaVLaS (approved):
/lgtm

PR #449, opened 2021-06-28
Fix and organize Hive Metastore descriptors from Trino

rimolive:


anishasthana (commented):
/lgtm

rimolive:
Most of the changes are just nit things to make the configuration files readable, but one small change will make hive metastore write their .db files locally instead of the S3 bucket.

crobby (approved):
lgtm, thanks for having an env where I could easily test this.

PR #454, opened 2021-07-08
Update superset deployment image

4n4nd:
This will update the Superset Deployment image to `quay.io/opendatahub/superset:1.1.0`, which has the drivers for superset to be able to connect with Trino as a db.


LaVLaS:
@maulikjs Is there any reason why `superset:1.1.0-trino` should not be the default superset image?

maulikjs:
@LaVLaS no reason, we actually just pushed this image recently and then added trino so this image exists, 1.2.0 onwards `trino` will be part of the default image.

superset/base/params.env
```diff
@@ -16,4 +16,4 @@ SUPERSET_DB_MEMORY_REQUESTS=300Mi
 SUPERSET_DB_MEMORY_LIMITS=1Gi
 SUPERSET_DB_CPU_REQUESTS=300m
 SUPERSET_DB_CPU_LIMITS=1
-SUPERSET_IMAGE=quay.io/internaldatahub/superset:1.1.0
+SUPERSET_IMAGE=quay.io/internaldatahub/superset:1.1.0-trino
```
anishasthana, line 19: Should this image reside in the opendatahub org on Quay?

4n4nd: @maulikjs @LaVLaS

rimolive: `opendatahub/superset` is empty, so I decided to keep the version tag without the additional name: 

```
quay.io/opendatahub/superset:1.1.0
```

@4n4nd change the code to point to this image

4n4nd: does it have the trino driver?

rimolive: Yes, I pulled from the `1.1.0-trino` in internaldatahub org to there. I just didn't use the `-trino` suffix.

4n4nd: updated :+1:

anishasthana:
It's deployed on the OPF cluster, changes look good.

4n4nd:
@rimolive is `quay.io/internaldatahub/superset:1.1.0-trino` the same as `quay.io/opendatahub/superset:1.1.0`? If so, we have `quay.io/internaldatahub/superset:1.1.0-trino` deployed in OperateFirst.

maulikjs (approved):

anishasthana (approved):

PR #457, opened 2021-07-12
Add Jupyterhub idle culler

mroman-redhat:
Added the JupyterHub Idle Culler. It looks for a secret generated by jupyterhub_config.py


jupyterhub/jupyterhub-idle-culler/base/buildconfig.yaml
```diff
@@ -0,0 +1,17 @@
+kind: BuildConfig
```
vpavlin, line 1: We should create a repository for this with proper requirements or Pipfile and pre-build the image instead of building it on the fly

jupyterhub/jupyterhub-idle-culler/base/deploymentconfig.yaml
```diff
@@ -0,0 +1,38 @@
...
+                  key: token
+          ports:
+            - containerPort: 8080
+          command:
+            - jupyterhub-idle-culler
+            - --timeout=259200  # 3 days
+            - --url=jupyterhub:8080/hub/api
+  triggers:
+    - type: ConfigChange
+    - type: ImageChange
```
vpavlin, line 31: We will not need image change if we pre-build the image

jupyterhub/jupyterhub-idle-culler/base/deploymentconfig.yaml
```diff
@@ -0,0 +1,38 @@
...
+        - name: jupyterhub-idle-culler
+          image: jupyterhub-idle-culler:latest
+          env:
+            - name: JUPYTERHUB_API_TOKEN
+              valueFrom:
+                secretKeyRef:
+                  name: jupyterhub-api-token
+                  key: token
+          ports:
+            - containerPort: 8080
```
vpavlin, line 24: I don't think we need to expose a port since it does not have any API endpoint

OR is it for metrics?

jupyterhub/jupyterhub-idle-culler/base/imagestream.yaml
```diff
@@ -0,0 +1,7 @@
+kind: ImageStream
```
vpavlin, line 1: No need for IS if we'd have a pre-build image

jupyterhub/jupyterhub-idle-culler/base/kustomization.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
```
vpavlin, line 1: Since this is kind of an "addon" to JH, it would make sense to me to make it into an overlay - WDYT?

4n4nd:
@vpavlin @vpavlin wouldn't we be able to add this service in the [jupyterhub-config](https://github.com/opendatahub-io/odh-manifests/blob/master/jupyterhub/jupyterhub/base/jupyterhub-configmap.yaml#L8) and run it as a JH service ([Docs](https://github.com/jupyterhub/jupyterhub-idle-culler#as-a-hub-managed-service))?

vpavlin:
IIRC there was a discussion about  whether we want culler in JH container or not and the conclusion was that 

1. From SRE perspective it is better to have it as a separate service 
2. We have environments where culling makes sense and where it does not make sense, so this way we can controll whether culler is enabled by modifying the CR (which is obviously also possible in JH Config, but this is more in context of 1.)

WDYT?

jupyterhub/jupyterhub-idle-culler/base/deploymentconfig.yaml
mroman-redhat: I can remove it now, and we can add it later if we need it.

4n4nd:
> From SRE perspective it is better to have it as a separate service

Yeah I realize that now. 
If the culler process breaks for some reason it might end up taking JH down as well. This way the culler stays independent of the JH server if anything breaks in the culler, it wouldn't affect the JH service.

Thanks!

jupyterhub/jupyterhub/overlays/jupyterhub-idle-culler/deploymentconfig.yaml
```diff
@@ -0,0 +1,28 @@
...
+    app: jupyterhub-idle-culler
+  replicas: 1
+  template:
+    metadata:
+      labels:
+        app: jupyterhub-idle-culler
+    spec:
+      containers:
+        - name: jupyterhub-idle-culler
+          image: quay.io/odh-jupyterhub/jupyterhub-idle-culler:latest
```
vpavlin, line 16: We will want to make a release in the idle culler image repo and point to that released image version here

jupyterhub/jupyterhub/overlays/jupyterhub-idle-culler/deploymentconfig.yaml
```diff
@@ -0,0 +1,28 @@
...
+          env:
+            - name: JUPYTERHUB_API_TOKEN
+              valueFrom:
+                secretKeyRef:
+                  name: jupyterhub-api-token
+                  key: token
+          command:
+            - jupyterhub-idle-culler
+            - --timeout=259200  # 3 days
```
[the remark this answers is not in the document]
4n4nd: we use 3 days in operate-first and haven't had a problem yet

jupyterhub/jupyterhub/overlays/jupyterhub-idle-culler/deploymentconfig.yaml
4n4nd: can we not just add the JH culler package to the [jupyterhub image pipfile](https://github.com/opendatahub-io/jupyterhub-odh/blob/master/Pipfile)? This would mean one less image for us to maintain and this would also mean that users would have the option to use it as a JH service if they wanted.

jupyterhub/jupyterhub/overlays/jupyterhub-idle-culler/deploymentconfig.yaml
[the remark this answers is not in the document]
vpavlin: ok, we might want to parameterize this in the future - either via param or a CM/env var value (just a note)

jupyterhub/jupyterhub/overlays/jupyterhub-idle-culler/deploymentconfig.yaml
```diff
@@ -0,0 +1,28 @@
...
+    app: jupyterhub-idle-culler
+  replicas: 1
+  template:
+    metadata:
+      labels:
+        app: jupyterhub-idle-culler
+    spec:
+      containers:
+        - name: jupyterhub-idle-culler
+          image: quay.io/odh-jupyterhub/jupyterhub-img:v0.3.2
```
mroman-redhat, line 16: I set img version to 0.3.2 in anticipation of a future update which will include the change on the jupyterhub-odh side

jupyterhub/jupyterhub/overlays/jupyterhub-idle-culler/deploymentconfig.yaml
```diff
@@ -0,0 +1,28 @@
...
+          env:
+            - name: JUPYTERHUB_API_TOKEN
+              valueFrom:
+                secretKeyRef:
+                  name: jupyterhub-idle-culler
+                  key: token
+          command:
+            - jupyterhub-idle-culler
+            - --timeout=259200  # 3 days
+            - --url=jupyterhub:8080/hub/api
```
vpavlin, line 26: @mroman-redhat The port of JH API changed to 8081, can you please change it?

Can you also rebase?

vpavlin:
lgtm

vpavlin (approved):

PR #459, opened 2021-07-13
Refactoring on Superset manifests

rimolive:


superset/README.md
```diff
@@ -11,17 +11,73 @@ To install Superset add the following to the `kfctl` yaml file.
...
+#### storage_class
+
+Name of the storage class to be used for PVC created by Hue's database. This requires `storage-class` **overlay** to be enabled as well to work.
+#### superset_secret
+
+The secret containing the environment variables to set the admin credentials and the Superset app secret key used to encrypt information in the database. If not specified, environment variables from [`superset`](base/secret-superset.yaml) will be used.
+
+When creating a custom secret, the following information should be added:
+
+* **SUPERSET_ADMIN_USER**: The username of the Superset administrator. If using the `superset` secret, `admin` will be used
```
anishasthana, line 35: Should this instead say "If not using the `superset` secret"?

superset/base/kustomization.yaml
```diff
@@ -25,143 +24,94 @@ configMapGenerator:
...
 
 configurations:
   - params.yaml
 
 generatorOptions:
   disableNameSuffixHash: true
+
+images:
+  - name: superset
+    newName: quay.io/internaldatahub/superset
```
anishasthana, line 113: Isn't there a different image we want to use here?
@maulikjs @4n4nd

anishasthana: Other relevant PR: https://github.com/opendatahub-io/odh-manifests/pull/454

rimolive: I changed to `quay.io/opendatahub/superset:1.1.0`

superset/README.md
rimolive: Actually, that's what I meant: If you use the default secret you will get these values set by default.

Do you think I should reword this?

anishasthana: I think I'd rephrase it to say "If using the default `superset` secret,"....

maulikjs (approved):

anishasthana (approved):

PR #462, opened 2021-07-20
Add required permissions for adding notebooks in custom namespaces

mroman-redhat:
This PR should add all the neccessary permissions to allow JupyterHub to spawn a notebook on a different namespace than the jupyterhub pod itself. Related to: https://github.com/opendatahub-io/jupyterhub-odh/pull/101

### DIFF
--- a/jupyterhub/jupyterhub/base/jupyterhub-cluster-role.yaml
+++ b/jupyterhub/jupyterhub/base/jupyterhub-cluster-role.yaml
@@ -6,6 +6,17 @@ metadata:
   name: jupyterhub-cluster
 rules:
   - verbs:
+      - create
+      - delete
+      - deletecollection
+      - patch
+      - update
+    apiGroups:
+      - ''
+    resources:
+      - persistentvolumeclaims
+      - pods
+  - verbs:
       - get
       - watch
       - list
@@ -22,3 +33,5 @@ rules:
       - ''
     resources:
       - nodes
+      - pods
+      - events
--- a/jupyterhub/jupyterhub/base/jupyterhub-cluster-rolebinding.yaml
+++ b/jupyterhub/jupyterhub/base/jupyterhub-cluster-rolebinding.yaml
@@ -1,7 +1,7 @@
 kind: ClusterRoleBinding
 apiVersion: rbac.authorization.k8s.io/v1
 metadata:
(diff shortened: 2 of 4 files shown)


jupyterhub/jupyterhub/base/jupyterhub-cluster-role.yaml
```diff
@@ -22,3 +33,5 @@ rules:
       - ''
     resources:
       - nodes
+      - pods
+      - events
```
anishasthana, line 37: Missing newline

anishasthana:
The following commands let us deploy to a second namespace

```
# To allow pod creation
oc adm policy add-scc-to-user nonroot -z default -n <notebook-namespace>
# To enable pulling images
oc policy add-role-to-user system:image-puller system:serviceaccount:<notebook-namespace>:default --namespace=<jupyterhub-namespace>
```

The additional privileges of nonroot don't seem that bad -- from the docs:

```
nonroot provides all features of the restricted SCC but allows users to run with any non-root UID. The user must specify the UID or it must be specified on the manifest of the container runtime.
```

crobby:
> The following commands let us deploy to a second namespace
> 
> ```
> # To allow pod creation
> oc adm policy add-scc-to-user nonroot -z default -n <notebook-namespace>
> # To enable pulling images
> oc policy add-role-to-user system:image-puller system:serviceaccount:<notebook-namespace>:default --namespace=<jupyterhub-namespace>
> ```
> 
> The additional privileges of nonroot don't seem that bad -- from the openshift blog[1]:
> 
> ```
> nonroot provides all features of the restricted SCC but allows users to run with any non-root UID. The user must specify the UID or it must be specified on the manifest of the container runtime.
> ```
> 
> [1] https://www.openshift.com/blog/managing-sccs-in-openshift

Are you saying that we'd need the extra scc/rolebinding for this PR to work?

anishasthana:
@crobby We'd need both extra role bindings for jupyterhub to be able to deploy notebooks to another namespace.

anishasthana:
@mroman-redhat I don't think this PR is needed any more: https://github.com/opendatahub-io/odh-manifests/pull/465 should cover the required rbac

nakfour:
@mroman-redhat please delete this PR since Anish's PR is the right solution

PR #463, opened 2021-07-20
WIP: Add custom namespace overlay

anishasthana:


anishasthana:
The issue I am running into right now is that the generated yaml artifacts have the wrong value for the rolebinding's `subjects/namespace`. Even though I am specifying `$(destination)` in the value for namespace, it is overwritten by `metadata.namespace`. I'm not sure how we can have an overlay given this problem.

anishasthana:
https://github.com/kubernetes-sigs/kustomize/issues/880 seems to be related.

anishasthana:
https://github.com/opendatahub-io/odh-manifests/pull/465 replaces this

PR #472, opened 2021-08-11
Use k8s jobs for cuda build chains deployment

LaVLaS:
Move the deployment of the cuda buildchains to a kubernetes job so that they can be deployed once and allow the user to control the redeployment of any failed builds

You can test deployment of the `cuda-11-0-3-build` job using this KfDef 
[pr-142.kfdef.txt](https://github.com/opendatahub-io/odh-manifests/files/7443081/pr-142.kfdef.txt)

Deletion of of the cuda-11.0.3 objects can be confirmed with the steps outlined in the PR [README](https://github.com/opendatahub-io/odh-manifests/blob/ce2a5d5fa0c54736ff5c4150703e25066beb66f6/jupyterhub/notebook-images/overlays/cuda-11.0.3/README.md#deleting-cuda-build-objects)

Closes https://issues.redhat.com/browse/ODH-438


jupyterhub/notebook-images/overlays/cuda-11.0.3/cuda-build-job.yaml
```diff
@@ -0,0 +1,50 @@
...
+                  fieldPath:
+                    metadata.namespace
+
+          command:
+            - /bin/bash
+            - -c
+            - |
+              set -x
+              echo "PWD: $PWD"
+              oc create -n ${BUILD_NAMESPACE} -f /tmp/gpu-notebook.yaml
```
vpavlin, line 32: OpenShift recommends using `apply` rather than `create`

LaVLaS: I am using `oc create` to prevent any situation where this job would re-apply or reset the BuildConfigs and ImageStream when they already exist.

vpavlin: Doesn't that prevent us from doing upgrades?

vpavlin: A job only runs once (unless it fails), no matter how many times you re-apply it (IIRC), so it does not really matter. But when we want to update the chain, we create a new job and the job would fail as it would not be able to create the resources again.

nakfour: Please open an issue for the re-design or re-factoring of cuda builds and version numbers.

LaVLaS:
/unhold

The image pull authorization issues have been resolved with the change to `jupyterhub-notebook-image-puller-rb.yaml`

PR #473, opened 2021-08-24
Bugfix JupyterHub Leader Election

lucferbux:
---
Title: Bugfix JupyterHub Leader Election
---

### Changes proposed in this pull request
As we have implemented High Availability on Jupyterhub, we shifted from 1 to 3 containers with the Leader Election strategy.

With this implementation, we could bump into inconsistencies given that one Pod could be thinking he's still the leader elected, but others could have replaced it due to network problems.

Implement a check to detect if a pod running is still the leader elected, and if not, delete it.

#### Diagram

leader-election

### Issues

* https://issues.redhat.com/browse/ODH-453

### Test instructions

* Check the instructions inside the [testing documentation](https://docs.google.com/document/d/1Ga0khZqr-jGpi5ExPJ0Cz8RawI0INc0UGI6Q2O3fTJc/edit#) about the ODH Manifest deployment
* Use this branch as the base image
* You can use olminstall to deploy to the dev cluster the changes
[TODO: How to check that changes are working]

### DIFF
--- a/jupyterhub/jupyterhub/base/jupyterhub-dc.yaml
+++ b/jupyterhub/jupyterhub/base/jupyterhub-dc.yaml
@@ -141,7 +141,7 @@ spec:
             secretKeyRef:
               name: $(traefik_credentials_secret)
               key: TRAEFIK_API_PASSWORD
-        image: quay.io/odh-jupyterhub/jupyterhub-img:v0.3.3
+        image: quay.io/odh-jupyterhub/jupyterhub-img:v0.3.4
         imagePullPolicy: "Always"
         name: jupyterhub
         ports:
@@ -162,6 +162,14 @@ spec:
               port: 8081
             initialDelaySeconds: 15
             periodSeconds: 10
+        livenessProbe:
+          exec:
+            command:
+            - /bin/sh
+            - -c
+            - /opt/app-root/src/readinessProbe.sh
+          initialDelaySeconds: 40
+          periodSeconds: 10
       - name: jupyterhub-ha-sidecar
         image: 'registry.redhat.io/openshift4/ose-leader-elector-rhel8:v4.7'
         args:
@@ -180,6 +188,41 @@ spec:
               fieldRef:
                 apiVersion: v1
                 fieldPath: metadata.name
+        livenessProbe:
+          exec:
+            command:
+            - /bin/sh
+            - -c
+            - >
+              function is_leader_election_ready()() {
+                  [ $(curl -s -o /dev/null -w ''%{http_code}'' localhost:4040) == "200" ]
+              }
+
+              function get_leader() {
+                  echo "$(curl http://localhost:4040 2> /dev/null | python3 -c "import sys, json; print(json.load(sys.stdin)['name'])")"
+              }
+
+              function is_empty_string() {
+                  # When the sidecar container is not working the name of the leader is an empty string
+                  [ "$(get_leader)" == "" ]
+              };
+
+              if ! is_leader_election_ready(); then
+                  # Sidecar container is not ready yet
+                  exit 0
+              fi
+
+              if is_empty_string; then
+                  # Sidecar container is returning an empty string, there must be a problem
+                  # Sleep for a while to let the rest of the container to restart too
+                  sleep 9
+                  exit 1
+              else
+                  # Pod is working, we can exit with success
+                  exit 0
+              fi
+          initialDelaySeconds: 40
+          periodSeconds: 10
       initContainers:
       - command:
         - wait-for-database
@@ -191,7 +234,7 @@ spec:
               key: POSTGRESQL_PASSWORD
         - name: JUPYTERHUB_DATABASE_HOST
           value: jupyterhub-db
-        image: quay.io/odh-jupyterhub/jupyterhub-img:v0.3.3
+        image: quay.io/odh-jupyterhub/jupyterhub-img:v0.3.4
         imagePullPolicy: "Always"
         name: wait-for-database
         resources:


lucferbux:
I have found a weir issue, those two images that contain the same s2i code behave different with the leader election strategy:

- quay.io/odh-jupyterhub/jupyterhub-img:v0.3.3 -> Works as intended
- quay.io/modh/odh-jupyterhub-container:v1.1.1-10 -> Doesn't work at all (even though it has the s2i run script that determine the leader election strategy)

lucferbux:
> Is it just me or are the commits for this PR weird?

Yeah I did a last minute change cause I didn't know if it was targeted for the code free, I'm refactoring and updating the PR.

lucferbux:
> @lucferbux Can you rebase to pull in the fix for the JupyterHub CI failures?

Great thanks! Done!

lucferbux:
@LaVLaS I still have issues with the tests. I think the problem lies in selenium, take a look:

```
SessionNotCreatedException: Message: session not created: This version of ChromeDriver only supports Chrome version 91
Current browser version is 93.0.4577.63 with binary path /usr/bin/google-chrome
```

I've bumped into this error in other projects, and it's usually caused by selenium updating the chrome version but the host machine having a prior ChromeDriver. 

I'm not a 100% sure if that's the cause, but if you want I can take a look in that direction!

vpavlin:
> I've bumped into this error in other projects, and it's usually caused by selenium updating the chrome version but the host machine having a prior ChromeDriver.

This is due to some updates in selenium vs. Chrome Driver

Update the `CHROMEDRIVER_VER` in https://github.com/opendatahub-io/odh-manifests/blob/master/tests/Dockerfile to the value in the error message and it should start working again

lucferbux:
> > I've bumped into this error in other projects, and it's usually caused by selenium updating the chrome version but the host machine having a prior ChromeDriver.
> 
> This is due to some updates in selenium vs. Chrome Driver
> 
> Update the `CHROMEDRIVER_VER` in https://github.com/opendatahub-io/odh-manifests/blob/master/tests/Dockerfile to the value in the error message and it should start working again

Great thanks @vpavlin trying now with new version!

lucferbux:
Great, is now working with the fix proposed, i would need aprooval.
@anishasthana @maulikjs @gmfrasca @rimolive

vpavlin:
> Thanks @vpavlin i was thinking about that, this is just the preliminary PR in order to test the issue.

@lucferbux What about my comment that it should not be implemented directly in yaml?:-)

lucferbux:
> > Thanks @vpavlin i was thinking about that, this is just the preliminary PR in order to test the issue.
> 
> @lucferbux What about my comment that it should not be implemented directly in yaml?:-)

Oh yeah, sorry I totally forgot to answer you 🤦‍♂️. OK, first of all, i removed the script in the sidecar container cause it wasn't the right solution, and second, yeah I totally agree, it could be a great solution although there are a few drawbacks.

1. Having script referenced in the yaml make harder to check the functionality of a container at first glance, I'm totally ok with that but it's a thing to consider.
2. Sorry for being a n00b, but I'm not 100% sure how to implement scripts outside yaml rather than copy them in the images and then run them in the location, if there's an easier solution, please let me know @vpavlin 

But yeah, sorry cause I discuss it with someone else but I didn't come back to you, thanks for the reminder!

vpavlin:
No worries:-) 


> 1. Having script referenced in the yaml make harder to check the functionality of a container at first glance, I'm totally ok with that but it's a thing to consider.

I am not sure what you mean by that - the reason for this to be called by livenessProbe is to not have to check "visually", but let the cluster handle the liveness of the pods


> 2. Sorry for being a n00b, but I'm not 100% sure how to implement scripts outside yaml rather than copy them in the images and then run them in the location, if there's an easier solution, please let me know @vpavlin

No other way - only way it to put the script in the image and call it from `livenessProbe`, but it's probably better than having shell code in YAML
> 
> But yeah, sorry cause I discuss it with someone else but I didn't come back to you, thanks for the reminder

lucferbux:
> No worries:-)
> 
> > 1. Having script referenced in the yaml make harder to check the functionality of a container at first glance, I'm totally ok with that but it's a thing to consider.
> 
> I am not sure what you mean by that - the reason for this to be called by livenessProbe is to not have to check "visually", but let the cluster handle the liveness of the pods
> 
> > 1. Sorry for being a n00b, but I'm not 100% sure how to implement scripts outside yaml rather than copy them in the images and then run them in the location, if there's an easier solution, please let me know @vpavlin
> 
> No other way - only way it to put the script in the image and call it from `livenessProbe`, but it's probably better than having shell code in YAML
> 
> > But yeah, sorry cause I discuss it with someone else but I didn't come back to you, thanks for the reminder

Perfect! So I will definitely do that, although, I'm gonna cherry pick the fix testing commit into another PR in order to unblock that path, cause I think @rimolive needs that ASAP.

vpavlin:
Yeah, I think it is ok if we merge this for now if it is necessary, but we should definitely try to get it done "right" ASAP to not introduce techical debt

lucferbux:
> Yeah, I think it is ok if we merge this for now if it is necessary, but we should definitely try to get it done "right" ASAP to not introduce techical debt

Oh no no, no worries, I can work on that and have the changes in this same PR, the "urgent" part was fixing the tests, so I'm working on that right now, thanks again @vpavlin

vpavlin:
> the "urgent" part was fixing the tests,

ah, gotcha

lucferbux:
I've added a 9-second sleep in the liveness probe of the sidecar container, cause I've been testing the changes in my dev cluster and the error triggers in both 3 sidecar containers at the same time, but if their liveness probe are not synced, some of them will be restarted, and hence they will work again and some of them not and they'll keep malfunctioning

lucferbux:
Btw, this PR should not be merged *until* this other PR is approved and there's a new jupyterhub-odh img to link.
https://github.com/opendatahub-io/jupyterhub-odh/pull/116

vpavlin (changes requested):
Once the image is updated, I think we are good to merge

anishasthana (commented):
/lgtm

vpavlin (approved):

PR #481, opened 2021-09-20
Enable system:authenticated users as JH default user

LaVLaS:
Recent JH updates have disabled `rhods-users: ''` as an option for allowing all user access and this change will restore that functionality. 


LaVLaS:
I am closing this as we don't need to upstream

PR #485, opened 2021-09-30
Add a ui configmap for jsp and enable gpu

maroroman:
This configmap will add the AWS env var category and enable the GPU dropdown by default.


LaVLaS (approved):
/lgtm

I haven't tested this with an enabled gpu but this is the same configmap that was used by other users to restore GPU selection in the JH Spawner UI

PR #491, opened 2021-10-12
Increase the memory limit for the configmap-puller container

maulikjs:
https://issues.redhat.com/browse/ODH-505


anishasthana (approved):
/lgtm

PR #498, opened 2021-11-02
Adding an example kfdef for odh on OCP4.9

nakfour:
This PR is an example of an ODH kfdef that currently works on OCP 4.9


LaVLaS (approved):
/lgtm

PR #507, opened 2021-11-25
Ceph Dashboard is not accessible. 

Jooho:
Signed-off-by: Jooho Lee <ljhiyh@gmail.com>

Fix ceph dashboard access issue(ODH-557)


nakfour:
@Jooho should we just create a route to the dashboard instead of doing port forward. Also we need to add this to the ODH dashboard.

Jason4849:
@nakfour I thought like you at first time but with my testing, I realised that Route did not work.  I tested other several scenarios but I confirmed the only working way was `port-forward`. 

Let me explain more about my test.
- The statefulset does not have sree port so it is not possible to create a route for console. After I added the port in the statefulset and exposed the port with Route. However, the sree could not access to the ceph when I clicked the bucket because it does not use Route hostname. It used `127.0.0.1` 
- With ceph dashboard, it is also possible to control bucket but it needs more commands so I don't think it is user-friendly.

FYI, I wrote this simple docs for ceph-nano installation.
- [Ceph nano installation](https://github.com/Jooho/jhouse_openshift/blob/master/docs/ceph-nano/ceph-nano-installation-on-openshift.md)
- [Ceph dashboard enable](https://github.com/Jooho/jhouse_openshift/blob/master/docs/ceph-nano/ceph-dashboard-enabl.md)

LaVLaS (approved):
/lgtm

I was able to port-forward and access the dashbaord locally.  The webui was not 100% reliable when creating buckets but I was able to create, delete and upload files via the dashboard

Jooho:
@LaVLaS I merged master and looks good now.

PR #510, opened 2021-12-13
Configure Superset to use OpenShift OAuth

accorvin:
This change:

1. Configures Superset to use OpenShift OAuth for its identity provider
2. Switches the superset route to use SSL
3. Makes it easier to extend the default superset configuration set in superset_config.py

This addresses https://issues.redhat.com/browse/ODH-566

To test this:

1. Run the following commands:
```
oc new-project opendatahub
kustomize build superset/base/ | oc apply -f -
```
2. Navigate to Superset (get the URL from the Superset route)
3. Click the login button in the top right of Superset
4. Log in with OpenShift
5. Verify that you are a Superset admin (you should be able to see admin functionality (e.g. list users, roles) when you hover over the settings menu in the top right)
6. Have someone who isn't a project admin in your openshift cluster test access and confirm that they are not a superset admin


superset/README.md
```diff
@@ -77,7 +80,28 @@ This parameter will configure the CPU request for Superset Database. If not set,
 
 ### Superset config file customization
 
-Superset manifests comes with a [`superset-config`](base/secret.yaml) secret, which will configure basic parameters for Superset like the database SQLAlchemy URL to connect to its database.
+In addition to the above parameters, Superset is configured via a [`superset_config.py`](base/superset_config.py) script.
+This script configures a number of important options, including:
+  * The SQLAlchey URL for Superset's Database
```
gmfrasca, line 85: SQLAlchemy

superset/base/kustomization.yaml
```diff
@@ -23,6 +26,11 @@ configMapGenerator:
     envs:
       - params.env
 
+secretGenerator:
```
HumairAK, line 29: does it have to be a secret? seems like everything confidential is pulled in via env vars, so could be a config map no?

accorvin: I think that's a great idea. It's something I was thinking about too. Our deployment doesn't require it being a secret. If the op-first one doesn't either then I think a configmap would be good. I'll work on changing it.

accorvin: Done.

accorvin:
One thing to be aware of with this change: Upon enabling it, existing Superset deployments are likely to have conflicts/issues when changing from an existing IdP to OpenShift oauth, specifically around username collisions. These deployments should plan to extend superset_config.py (this PR makes it easy to do so) to add logic to handle this.

LaVLaS (approved):
/lgtm

The rollout of superset worked from me and no issues with the login with my OCP oauth credentials

PR #512, opened 2021-12-17
Add pod resources to the traefik initContainer

LaVLaS:
Traefik initContainer was missing defined resources that could cause issues in clusters that implement resource quotas


anishasthana (approved):
These are identical to the limits we've had in RHODS for a while (which have worked fantastically).

/lgtm

PR #513, opened 2021-12-22
Remove unused superset-data PVC

accorvin:
This PVC was previously used when Superset used a local SQLite database.
Since switching to an external postgres database this PVC is no longer
used, so this commit deletes it.

This addresses https://issues.redhat.com/browse/ODH-570


LaVLaS (approved):
/lgtm

PR #515, opened 2022-01-06
Update superset image to 1.3.2

LaVLaS:
Update the Superset component to v1.3.2 using the image built from https://github.com/amancevice/docker-superset

Fixes: https://issues.redhat.com/browse/ODH-541

Test Steps:
1. Login to superset as the user who is in the namespace admin group
2. Navigate to the database page - `Data -> Database`
3. Attempt to add a database `+ Database`
    You can use PostgreSQL superset-db pod for the test
    DB Name: `superset`
    User / Pass: `changeme` / `changeme`

The db should be created successfully


accorvin (approved):
Change looks good to me. Confirmed to be working as described in testing instructions. I also confirmed that upgrade from 1.3.0 to 1.3.2 works without issue.

PR #517, opened 2022-01-14
Fix certification verification issue for superset oauth

accorvin:
In the code that was added for Superset OAuth, we specify the path to
the certificate bundle to be used for verifying requests to the local
openshift cluster's oauth endpoint.

The cert that was used is the openshift cluster certificate. It turns
out that in cases when a globally trusted certificate is used (rather
than a cluster self-signed cert), this openshift cluster certificate
cannot be used to validate the oauth endpoint. Instead, in these cases
we must use the default system certificate bundle in /etc/ssl/certs.
Unfortunately, this certificate bundle does not work on clusters that
generate a self-signed cluster cert.

This change combines the default global cert bundle with the openshift
cluster cert bundle and writes them together into a new file. This
resulting cert bundle file is then used by superset to perform oauth and
works on both types of clusters.


accorvin:
To test this change:

1) Spin up two clusters - one with a self-signed cluster cert (such as with the OSIA tool), one with a globally trusted cert (such as in OSD)
2) Deploy superset onto both of these clusters from the version currently in master (you can do the deploy by simply running `kustomize build superset/base | oc apply -n opendatahub -f -` )
3) Confirm that superset oauth works on the self-signed cert cluster, but not the globally trusted cert cluster
4) Deploy this change
5) Confirm that auth works on both superset instances

accorvin:
I have two clusters currently spun up if anyone wants to use them for testing.

LaVLaS:
This appears to fix issues with users unable to authorized on clusters using registered SSL certs.  Authenticating with openshift/google oauth, htpasswd and kubeadmin were successful.

CI failures appear to be linked to a recent ods [update](https://github.com/red-hat-data-services/ods-ci/pull/125) that forces a check for a downstream version to check for UI versions

nakfour:
Tried this on OCP 4.9, it got stuck on the log in window with an error in the pod
```2022-01-17 19:33:31,465:ERROR:flask_appbuilder.security.views:Error returning OAuth user info: argument of type 'NoneType' is not iterable```

accorvin:
@nakfour can you tell me anything about the cluster you're testing on? How was it provisioned (e.g. RHPDS, OSIA, OSD)? What identity provider are you using?

nakfour:
@accorvin it is RHPDS with lets encrypt

accorvin:
@nakfour I've added a commit which fixes your issue.

anishasthana (commented):
/lgtm

LaVLaS (approved):
/lgtm

I worked when I tested against in-cluster openldap and corporate ldap identityProviders.  

Only issue I experienced was with `kubeadmin` where login was successful but logging out then clicking login would automatically login back without allowing the user to authenticate again.  The workaround was to clear the browser cache

PR #522, opened 2022-01-31
Remove startingCSV

strangiato:
Currently startingCSV's are manually set in the operator subscriptions.  This causes the operators to be installed with an older version of the operator and OLM will then slowly update the operator to the latest version of the operator available within that subscription channel.

When startingCSV is not set, OLM will select the startingCSV automatically and default to the latest version available within that subscription channel.

### DIFF
--- a/grafana/cluster/base/subscription.yaml
+++ b/grafana/cluster/base/subscription.yaml
@@ -9,4 +9,3 @@ spec:
   name: grafana-operator
   source: community-operators
   sourceNamespace: openshift-marketplace
-  startingCSV: grafana-operator.v3.9.0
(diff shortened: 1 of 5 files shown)


strangiato:
It looks like the Seldon operator Subscription sets a `startingCSV` to 1.7.0 which was removed from OperatorHub several days ago.  Instead, the subscription needs 1.12.0 to install successfully or removing the startingCSV would allow OLM to set the initial startingCSV.

LaVLaS:
This looks good. In the initial days of OperatorHub, our original purpose for setting `startingCSV` was to preserve compatibility between components when there was a dependency for specific versions.  Now that OperatorHub has matured and operators are updating much faster than ODH this makes sense.  Going forward, I would assume that this should be the default unless there is strong justification for pinning the starting version

crobby (approved):
/lgtm

PR #523, opened 2022-02-01
Update kfdef used for OCP 4.9

LaVLaS:
This update the kfdef that we provided specifically for OCP 4.9

After updating the kfdef repoRef to use `master`, you can test this on any OCP 4.9 cluster.


nakfour:
I tested this on 4.9 and it worked. /lgtm

PR #543, opened 2022-03-23
add part-of label

strangiato:
This PR adds the `app.kubernetes.io/part-of` in any kustomize file where `component.opendatahub.io/name` is currently utilized.  The `app.kubernetes.io/part-of` label is part of the recommended labels from kubernetes:

https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/

OpenShift utilizes the `app.kubernetes.io/part-of` label to group application resources in the Topology UI view:

image

This PR does not add the label to notebook pods spawned by JupyterHub which I believe will require changes to the jupyterhub_config.py file in the JupyterHub image.


LaVLaS:
Thank you.  This willl be really helpful to users who want to see how each component is connected in the OCP Developer console.  

Once this passes CI then we should be good to merge

strangiato:
@VaishnaviHire for the Operators the only object that we have the ability to label through the manifests is the Subscription object which as far as I know the Topology view doesn't utilize for the grouping.

I was testing on 4.8 so this may have changed in 4.9/4.10.

It probably wouldn't hurt to add the labels to the subscriptions in case the Topology view adds this as an enhancement though...

strangiato:
Rebased and added the labels to the subscription objects.

strangiato:
@VaishnaviHire I rebased from master again since the tests were failing and my branch wasn't up to date on the latest changes.

It looks like the tests are still continuing to fail for the check that Jupyterhub and Trino-worker pods have failed.

As far as I can tell none of the changes I have made should prevent the pods from starting or impact the `oc get` commands being utilizes in the tests.  I'm not seeing anything in the test logs to help indicate why the two pods are not starting.  Is there any way to get the events from the pods from the cluster that the tests are being performed on in order to get any additional information on why this is failing?

strangiato:
I figured out the issue with the tests and it looks like everything is passing now.

It looks like kustomize adds CommonLabels to both the objects and the selectors on the services.  On a Deployment it adds the labels to the pod template but it does not have the ability to add them to a DeploymentConfig so the service was not selecting the pods correctly.  The label had to be manually added to the pod template on the DC and that resolved the issue.

LaVLaS (approved):
/lgtm

PR #544, opened 2022-03-24
Upgrade superset to 1.4.1

rimolive:
This PR bumps Superset version to 1.4.1, and point to a UBI-based tag of this image.


LaVLaS:
Can you update the PR description to list what features are included in this update or why it is needed?

LaVLaS:
I need to confirm with a fresh deployment but I am unable to login successfully.  All attempts to login result in an infinite loop of from superset login -> openshift oauth login -> superset login. This is the only noticeable error in the logs
```
2022-03-25 22:26:26,059:ERROR:flask_appbuilder.security.views:Error returning OAuth user info: missing_token:
```

I deployed superset from the `master` branch with no issues

rimolive:
I think I know what's going on, and I made a change to a certificate path because in UBI the path is different than in the old community-based version. This doesn't explain the issue yet since I had a successful deployment on our stage cluster.

I'll review the changes that added OAuth support in Superset to understand what I need to do to fix it.

VaishnaviHire:
@rimolive @LaVLaS I see same login error with fresh deployment as well

rimolive:
@VaishnaviHire @LaVLaS The error was because the new superset image installed the latest version of authlib, which introduced this error. I fixed the authlib to 0.15.1 and it worked. You can test again.

The fix is in https://github.com/opendatahub-io/odh-images/pull/13

rimolive:
Tests now passed! waiting for reviews.

accorvin:
/reject
I think there is a regression in behavior here from the current 1.3.2 image. When you attempt to login via OAuth, you get sent to an additional superset screen with a button to "log in with openshift". In 1.3.2, you just go straight to the openshift auth screen when you click the initial superset login button.

I suspect that we've downgraded one of our pip packages in our custom build of the superset image relative to what the upstream superset community puts in their image.

rimolive:
From the latest Flask App-Builder docs: https://flask-appbuilder.readthedocs.io/en/latest/security.html

`On Flask-AppBuilder 3.4.0 the login page has changed.`

So probably I need to downgrade the fab version in the superset image (it uses 3.4.5 in the UBI-based image).

lucferbux (approved):
/lgtm

rimolive:
FWIW, there's an issue opened in superset repo too: https://github.com/apache/superset/issues/18573

The developers say it's not a bug, but a feature. :)

Issue #527, opened 2022-02-17
JupyterHub ServiceAccount permissions breaking automatically


vemonet:
**Describe the bug**
I deployed a JupyterHub with GitHub login a few weeks ago. It was working fine, I could start my notebooks

But when I came back a few weeks after without doing a single change to the project, I am now getting this error when trying to deploy a notebook:

```
 Error: HTTP 500: Internal Server Error (Error in Authenticator.pre_spawn_start: ForbiddenError 403 Reason: error HTTP response headers: HTTPHeaderDict({'Audit-Id': 'f832a95c-66f3-4a16-81f7-12dd46f4c085', 'Cache-Control': 'no-cache, private', 'Content-Type': 'application/json', 'X-Content-Type-Options': 'nosniff', 'X-Kubernetes-Pf-Flowschema-Uid': '1c8d3b08-c7cf-4e39-b24f-b3b0d783deab', 'X-Kubernetes-Pf-Prioritylevel-Uid': '502f2dec-904b-434b-aee7-87144b2ccead', 'Date': 'Thu, 17 Feb 2022 10:30:06 GMT', 'Content-Length': '283'}) HTTP response body: b'{"kind":"Status","apiVersion":"v1","metadata":{},
"status":"Failure","message":"nodes is forbidden: User \\"system:serviceaccount:github:jupyterhub-hub\\" cannot list resource \\"nodes\\" in API group \\"\\" at the cluster scope","reason":"Forbidden","details":{"kind":"nodes"},"code":403}\n' Original traceback: File "/opt/app-root/lib/python3.6/site-packages/openshift/dynamic/client.py", line 42, in inner resp = func(self, *args, **kwargs) File "/opt/app-root/lib/python3.6/site-packages/openshift/dynamic/client.py", 
line 247, in request _return_http_data_only=params.get('_return_http_data_only', True) File "/opt/app-root/lib/python3.6/site-packages/kubernetes/client/api_client.py", line 345, in call_api _preload_content, _request_timeout) File "/opt/app-root/lib/python3.6/site-packages/kubernetes/client/api_client.py", line 176, in __call_api _request_timeout=_request_timeout) File "/opt/app-root/lib/python3.6/site-packages/kubernetes/client/api_client.py", line 366, in request headers=headers) 
[...]

PR #545, opened 2022-03-25
make crb unique to each namespace

strangiato:
When multiple instances of JupyterHub are created in different namespaces on the same cluster, the ODH operator continuously updates the `jupyterhub-cluster` ClusterRoleBinding as it completes the reconciliation loop.

Since ClusterRoleBindings are cluster scoped and the name is not unique to each instance of ODH each instance of the KFDef is overwriting the same CRB to grant that specific service account permissions.

To verify the issue you can watch the `jupyterhub-cluster` CRB and as the ODH operator cycles through the reconciliation loop the service account namespace will swap back and forth between the different SA's created in the different namespaces.


strangiato:
Resolves #527

LaVLaS (changes requested):
Thank you for submitting this fix.  The will fail because the kfdef is not processing the `$(namespace)` parameter and is literally creating a `jupyterhub-cluster-$(namespace)` ClusterRolebinding.

I think you'll need to add an entry to the jupyterhub `params.yaml` to make it aware of  ClusterRolebinfing `metatadata/name`

PR #547, opened 2022-03-30
Update chromedriver to 100.0.4896.60

crobby:


crobby:
Ok, I think the tests were failing in this way before the chrome version update caused even more problems.  The chrome driver update appears to be working though, so I'm going to merge this.

PR #549, opened 2022-03-31
feat(byon): Update dashboard Role

tumido:
Support BYON backend when deployed to OCP.

This PR is the same as: https://github.com/opendatahub-io/odh-dashboard/pull/163

Question, why there's so many inconsistencies and different permissions in between

https://github.com/opendatahub-io/odh-manifests/blob/master/odh-dashboard/base/role.yaml

and

https://github.com/opendatahub-io/odh-dashboard/blob/master/install/odh/base/role.yaml

Shouldn't those manifests be the same?

/cc @anishasthana @cfchase @LaVLaS @crobby


LaVLaS:
/hold

> Question, why there's so many inconsistencies and different permissions in between
This is a good point.  The dashboard version that is provided in ODH is behind the master branch in the odh-dashboard repo.  Any updates to this manifest should be a bug fix in the manifest or paired with an official version update of the component in question.

I am putting this on hold since the role update does not appear to be tied to a bugfix for the version we deploy in odh-manifests

LaVLaS:
Closing this since it won't be included with the upcoming dashboard v2.0 release

PR #556, opened 2022-04-22
Add soft CPU affinity to the base configmap

Xaenalt:
JIRA https://issues.redhat.com/browse/RHODS-3074

Alternate solution to https://github.com/opendatahub-io/jupyterhub-odh/pull/135

This helps prevent the situation where a notebook that doesn't request GPUs can land on a GPU node unintentionally. In that case the notebook would have the ability to use the GPUs on that node, despite not requesting them. In the JIRA this is listed as a bit unintuitive for users, and possibly disruptive for some workloads.

The change causes notebooks to *prefer* CPU nodes, unless a GPU is requested. This will be honored if it can, though a notebook can still overflow to GPU nodes rather than making the notebook unschedulable. If that behavior is desired, the node taint workflow handles that case.


crobby:
So, if we merge this one, we won't need https://github.com/opendatahub-io/jupyterhub-odh/pull/135   correct?

Xaenalt:
Correct, however there is one concern between this and https://github.com/opendatahub-io/jupyterhub-odh/pull/135 in that it doesn't have a function to dynamically enable/disable the scheduling priority based on whether or not the user requested GPUs. That's not a problem from a functionality standpoint, but could look odd to a user inspecting the yaml. They'd see a GPU requested, and a soft anti-affinity rule for all GPU nodes.

Xaenalt:
@anishasthana @crobby @LaVLaS can we get a review on this and https://github.com/opendatahub-io/jupyterhub-odh/pull/135

PR #560, opened 2022-05-11
KServe raw deployment in ODH

crobby:
KServe Raw (without istio or cert-manager)


crobby:
Hey reviewers, can I get some reviews on this now.  You'll note that it's targeted at the kserve-prototype branch, so it's NOT for "real" yet.  I believe @DharmitD will be working on some tests for this and it will be nice to have a place for multiple people to work against to keep the commits smaller in the long-run.
The idea here was to take the stock upstream yamls and modify them minimally so that they'd work for us.  The only changes form upstream are:
1)  Remove "Issuer" and "Certificate" objects....they are only needed when cert-manager is used.  We are not using that, but instead relying on openshift to handle that for us
2) Adding some annotations so that openshift will handle the cert stuff (from above) for us

anishasthana:
I think we should test only on 4.10 or greater. 4.8 is already only on maintenance support
image

VaishnaviHire:
> I think we should test only on 4.10 or greater. 4.8 is already only on maintenance support image

Agreed. I just wanted to confirm if `kserve` manifests are updated to support 4.10. Looking at the changes,  they seem to be updated.

VaishnaviHire:
/lgtm

Successfully deployed sample Inferenceservice

PR #561, opened 2022-05-13
Add dashboard config CRD

maroroman:
Adds a odh-dashboard custom resource definition (and a default instance). Currently it replaces the values in the Dashboard configmap and adds notebook sizes, notebook controller configuration and notebook controller status.

The notebook controller part of the CR is supposed to essentially fill the role of the Jupyterhub Singleuser Profiles configmap, but for the Notebook controller.

The notebook controller status should be an array with the information about the user's previous choices and environment variables and a link to their envVar secrets (managed by the spawner UI)

Note: I do not know whether to also add other parts of the old JSP configmap, such as gpuTypes and profiles, although these can be added in together with the features that use them.

Related to: https://github.com/opendatahub-io/odh-dashboard/issues/159


maroroman:
@anishasthana @VaishnaviHire This CR has a problem that it would be reconciled by the operator, I am considering creating it in a init container of the dashboard when it is deployed to work around it, but let me know what would be the best solution.

odh-dashboard/base/odh-dashboard-crd.yaml
```diff
@@ -0,0 +1,97 @@
...
+                - notebookSizes
+              properties:
+                dashboardConfig:
+                  type: object
+                  properties:
+                    enablement:
+                      type: boolean
+                    disableInfo:
+                      type: boolean
+                    disableSupport:
```
dlabaj, line 32: Can you add these properties to the dashboardConfig?

 disableClusterManager: 
      type: boolean
 disableTracking: 
      type:boolean
  disableBYONImageStream: 
      type: boolean
  disableISVBadges: 
    type: boolean

dlabaj (changes requested):

LaVLaS:
@maroroman @dlabaj Will this be supported in the dashboard v2.0 release? If not, then we should put this on hold until it is included in a planned release

lucferbux (approved):
/lgtm

LaVLaS:
/hold

Putting this on hold until https://github.com/opendatahub-io/odh-dashboard/pull/238 is finalized and merged

odh-dashboard/base/odh-dashboard-crd.yaml
```diff
@@ -0,0 +1,103 @@
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  name: odhdashboards.opendatahub.io
+spec:
+  group: opendatahub.io
+  scope: Namespaced
+  names:
+    plural: odhdashboards
+    singular: odhdashboard
```
cfchase, line 10: this name makes me think you're creating an OdhDashboard and it's possible to create multiple dashboards.  I think we might want to call it an OdhDashboardConfig.

maroroman: Changed the name, I will also change the expected value in https://github.com/opendatahub-io/odh-dashboard/pull/238

LaVLaS (approved):
/lgtm

LaVLaS:
/unhold

This is required for the Dashboard v2.0 release

dlabaj (approved):
lgtm

PR #568, opened 2022-06-02
Jupyterhub Idle Culler update

maroroman:
This PR cherry-picks updates from downstream, as well as adds a value to the jupyterhub-cfg

Additionally it adds the content of PR https://github.com/red-hat-data-services/odh-manifests/pull/220 which adds limits and requests to the culler container.

This PR should also allow the dashboard to work with the culler, although the value change will not be applied unless the operator is scaled down.

### DIFF
--- a/jupyterhub/jupyterhub/base/jupyterhub-configmap.yaml
+++ b/jupyterhub/jupyterhub/base/jupyterhub-configmap.yaml
@@ -11,6 +11,7 @@ data:
   gpu_mode: ""
   singleuser_pvc_size: 2Gi
   notebook_destination: "$(notebook_destination)"
+  culler_timeout: "31536000" # 1 year in seconds
 
 ---
 apiVersion: v1
--- a/jupyterhub/jupyterhub/overlays/jupyterhub-idle-culler/deploymentconfig.yaml
+++ b/jupyterhub/jupyterhub/overlays/jupyterhub-idle-culler/deploymentconfig.yaml
@@ -42,6 +42,13 @@ spec:
                 configMapKeyRef:
                   name: jupyterhub-cfg
                   key: culler_timeout
+          resources:
+            limits:
+              cpu: 150m
+              memory: 50Mi
+            requests:
+              cpu: 150m
+              memory: 50Mi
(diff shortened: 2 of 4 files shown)


LaVLaS:
@maroroman Can you add a quick note to the JupyterHub README about enabling this overlay and how it's configured?

jupyterhub/jupyterhub/base/jupyterhub-configmap.yaml
```diff
@@ -11,6 +11,7 @@ data:
   gpu_mode: ""
   singleuser_pvc_size: 2Gi
   notebook_destination: "$(notebook_destination)"
+  culler_timeout: "31536000" # 1 year in seconds
```
LaVLaS, line 14: Does this need to be a kfdef parameter so that user's can configure this from the kfdef `kustomizeConfig`?

LaVLaS: Since ODH JH is planned for deprecation and given the adiditonal work required to allow the user to customize this configmap, we can merge this as is for now.

LaVLaS (approved):
/lgtm

Culler pod was restarted when the `culler_timeout` value in jupyterhub-cfg was changed

jupyterhub/jupyterhub/overlays/jupyterhub-idle-culler/deploymentconfig.yaml
```diff
@@ -11,6 +11,23 @@ spec:
       labels:
         app: jupyterhub-idle-culler
     spec:
+      initContainers:
+        - name: check-for-jh
+          image: 'quay.io/odh-jupyterhub/jupyterhub-img:v0.3.3'
```
dol, line 16: `image: 'quay.io/odh-jupyterhub/jupyterhub-img:v0.3.5'`

Should be increased to 0.3.5 like in this PR:  https://github.com/opendatahub-io/odh-manifests/commit/1fb9f35a934b5d8c9c92b7dc5c4488afd8d778b3

PR #569, opened 2022-06-02
Adding Kserve-raw tests

DharmitD:


crobby (changes requested):
I don't think we need the kserve.yaml file at all since the installation is handled by the manifests/operator

crobby (changes requested):
Looks like kserve-runtimes.yaml is also not needed (installed by operator/manifests).

DharmitD:
> Looks like kserve-runtimes.yaml is also not needed (installed by operator/manifests).

Done, updated to take both the files out.

DharmitD:
Force pushed to the commit to remove trailing whitespaces.

anishasthana:
Is this no longer needed as we are now going with Model Mesh?

DharmitD:
> Is this no longer needed as we are now going with Model Mesh?

I believe the plan is to have this merged into the prototype branch instead of master for now, and then eventually decide whether we need this or not. I'm okay with whatever, can keep these changed saved on my fork if anything. I'll let @crobby take the call.

crobby:
I've seen these tests run. I suspect the problem is due to our ci not being build to expect being ran from branches that are not master.  I'm inclined to merge this (well, once the changes are pared back down to the actual 6 or so files needed for the change) since it's just for the prototype branch--which isn't going to be needed for "awhile".

DharmitD:
> The changes should only be for about 6 files (approx). No need to go crazy on the whitespace for the whole project here.

Done, updated the commit to have only the files it had originally, along with whitespaces removed. Took out all the extra files.

crobby (approved):
lgtm.  I think the ci failure may be due to this running on a branch rather than master. Either way, I'm going to merge this since it's just for the prototype branch.

PR #571, opened 2022-06-13
Update Memory limit for Dashboard

VaishnaviHire:
This commits updates memory limits in Dashboard pod in order to reduce `OOMKill` errors


cfchase (approved):

LaVLaS (approved):
/lgtm

LaVLaS:
/test all

Running one last test just to see if there are still issues with CI test runs.  If it fails and it doesn't appear to be caused by this change then I'll merge it

PR #573, opened 2022-06-14
feat: Ray cluster deployment with ODH Jupyterhub

harshad16:
feat: Ray cluster deployment with ODH Jupyterhub

Description:
 - Inclusion of [Ray](https://docs.ray.io/en/latest/index.html) cluster setup using ODH Jupyterhub. This would give users to utilize the ray capabilities
 - Reference: https://old.operate-first.cloud/community-handbook/moc-ray-demo/README.md


jupyterhub/README.md
```diff
@@ -94,6 +94,10 @@ Contains build chain manifest for CUDA 11.0.3 enabled ubi 8 based images with py
 
 *NOTE:* Builds in this overlay require 4-6 GB of memory
 
+#### [odh-ray-cluster](notebook-images/overlays/odh-ray-cluster/)
+
+Contains deployment manifests for Ray setup with ODH jupyterhub.
```
anishasthana, line 99: I think with our move towrads kfnbc as the default notebook provider we should probably gear Ray towards nbc as well. @crobby  / @LaVLaS  wdyt?

jupyterhub/notebook-images/overlays/odh-ray-cluster/operator/ray-operator-deployment.yaml
```diff
@@ -0,0 +1,37 @@
...
+    metadata:
+      labels:
+        app: ray-operator
+    spec:
+      restartPolicy: Always
+      serviceAccountName: ray-operator
+      containers:
+        - name: ray-operator
+          imagePullPolicy: Always
+          image: 'quay.io/thoth-station/ray-operator:v0.1.2'
```
anishasthana, line 22: Should the image be in the Open Data Hub organization?

anishasthana: (even if it's just a mirror of the image in Thoth-station)

jupyterhub/notebook-images/overlays/odh-ray-cluster/ray-custom-resources.yaml
```diff
@@ -0,0 +1,4320 @@
+apiVersion: apiextensions.k8s.io/v1
```
anishasthana, line 1: So with this we're effectively going to continue requiring cluster admin for the ODH operator if you want to install Ray.

Would it instead be better to have these CRDs installed by the user before deployment of Ray artifacts?

anishasthana (commented):
_Should_ we be tying Ray in as an overlay for other components? 
I think scoping this to just deploying Ray would make more sense, no?

We can then create an overlay for deployment of ray custom images, or w/e

jupyterhub/notebook-images/overlays/odh-ray-cluster/ray-custom-resources.yaml
anishasthana: @LaVLaS I'm thinking about our previous discussion where we said ODH wants to get out of the business of installing operators, instead focusing on deploying _custom resources_ for already installed operators

jupyterhub/notebook-images/overlays/odh-ray-cluster/operator/ray-operator-deployment.yaml
harshad16: It just that, if we keep open-data-org image in here, people facing issue would have to look for the image in here.
i m happy to make the change, if that is way we want to go.

jupyterhub/notebook-images/overlays/odh-ray-cluster/ray-custom-resources.yaml
LaVLaS: ODH will no longer install operators that are available in OperatorHub. After doing a quick check, I don't see any OperatorHub packageManifest so this manual operator install would be the only option for now.

harshad16: yes, this is not available as the operator on operatorhub.

LaVLaS:
@harshad16 Thx for providing this.  Since we have plans to deprecate JupyterHub and migrate to KFNBC, can you separate the Ray operator into a standalone manifest so we can use the deployment with JH or KFNBC.

harshad16:
> Since we have plans to deprecate JupyterHub and migrate to KFNBC, can you separate the Ray operator into a standalone manifest so we can use the deployment with JH or KFNBC.

Thanks for the review @LaVLaS 
Should i move the ray operator manifest to the base? or some other director?

LaVLaS:
> > Since we have plans to deprecate JupyterHub and migrate to KFNBC, can you separate the Ray operator into a standalone manifest so we can use the deployment with JH or KFNBC.
> 
> Thanks for the review @LaVLaS Should i move the ray operator manifest to the base? or some other director?

The base repo will work.  

Also, is this ray-operator manifest customized in any way from the helm-chart to run on OCP?

LaVLaS (changes requested):
Before we start the official review, we'll need some basic tests to verify that it deploys successfully and a simple "hello world" test

jupyterhub/README.md
MichaelClifford: @anishasthana @LaVLaS 

Currently we are using a Single User Profile as part of our Ray /Jupyterhub deployment that spawns a Ray cluster for the user when they start the "ray notebook image" from the spawner page.  Is it possible to do something similar with KFNBC? 

Is there an analogues spawner page UI with KFNBC right now? or is it still done by `oc apply`?

MichaelClifford:
> Also, is this ray-operator manifest customized in any way from the helm-chart to run on OCP.

The operator deployment is slightly different then the helm-chart, but not sure if the changes were OCP specific or specific to the thoth images were using.  @erikerlandson would know.

harshad16:
Based on the conversation its seem we would need to make adjustment for ray to function on notebook controller and jupyterhub.
@MichaelClifford would be taking forwards this work in a different PR.
Thanks everyone for the help and thoughts :100:

PR #576, opened 2022-06-21
Fix parameter placement for hive-metastore resource requests and limits

rimolive:


VaishnaviHire:
I think we would also need to update the default values in `params.env` https://github.com/opendatahub-io/odh-manifests/blob/d2fbbe34f4b2bce3f0692b2133b016fb50d9d9e8/trino/base/params.env#L8

rimolive:
> I think we would also need to update the default values in `params.env`
> 
> https://github.com/opendatahub-io/odh-manifests/blob/d2fbbe34f4b2bce3f0692b2133b016fb50d9d9e8/trino/base/params.env#L8

I made the changes in README with the default values for each parameter. Are you suggesting other default values or just ensure the values in README are the same as in params.env?

Jooho:
It looks `trino_cpu_request` default value is also wrong. From README, `trino_cpu_request` default is 1 but it is actually 300m in params.env. If it is ok, could you update the param in the README file as well?

VaishnaviHire:
Worked well with ocp 4.10

/lgtm

PR #578, opened 2022-06-22
Updated liveness probe

lucferbux:
Updated liveness probe of jupyterhub


LaVLaS:
Closing this since it is no longer relevant

PR #584, opened 2022-07-04
Import notebook controller images using ImageStreams

samuelvl:
Fix a issue with the ODH notebook controller where the notebook couldn't start when the notebook images was referenced by a ImageStream.

See https://github.com/opendatahub-io/kubeflow/pull/27


LaVLaS (approved):
/lgtm

PR #613, opened 2022-08-02
Adding a monitoring check to pipeline tests

DharmitD:
## Description
- Added a monitoring check to pipeline tests
- Enabled user workload monitoring
- Updated the component name from `kfp-tekton` to `ml-pipelines` in the `kfctl_openshift.yaml` kfdef.

## Testing instructions

- Install the `OpenDataHub` and `OpenShift Pipelines` operators on the cluster.
- Create an `opendatahub` namespace: `oc new-project opendatahub`
- Run the following commands:
  - `make build -- to create the test container image`
  - `make run TESTS_REGEX=ml-pipelines`
- [X] The commits are squashed in a cohesive manner and have meaningful messages.
- [X] Testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).
- [X] The developer has manually tested the changes and verified that the changes work

### DIFF
--- a/tests/basictests/ml-pipelines.sh
+++ b/tests/basictests/ml-pipelines.sh
@@ -47,6 +47,17 @@ function check_run_status() {
     os::cmd::try_until_text "curl ${SVC}/apis/v1beta1/runs/${RUN_ID} | jq '.run.status'" "Completed" $odhdefaulttimeout $odhdefaultinterval
 }
 
+function setup_monitoring() {
+    header "Enabling User Workload Monitoring on the cluster"
+    oc apply -f ${RESOURCEDIR}/modelmesh/enable-uwm.yaml
+}
+
+function test_metrics() {
+    header "Checking metrics for total number of runs, should be 1 since we have spun up 1 run"
+    monitoring_token=`oc sa get-token prometheus-k8s -n openshift-monitoring`
+    os::cmd::try_until_text "oc -n openshift-monitoring exec -c prometheus prometheus-k8s-0 -- curl -k -H \"Authorization: Bearer $monitoring_token\" 'https://thanos-querier.openshift-monitoring.svc:9091/api/v1/query?query=run_server_run_count' | jq '.data.result[0].value[1]'" "1" $odhdefaulttimeout $odhdefaultinterval
+}
+
 function delete_runs() {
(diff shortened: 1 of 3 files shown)


tests/basictests/ml-pipelines.sh
```diff
@@ -47,6 +47,17 @@ function check_run_status() {
...
 }
 
+function setup_monitoring() {
+    header "Enabling User Workload Monitoring on the cluster"
+    oc apply -f ${RESOURCEDIR}/modelmesh/enable-uwm.yaml
+}
+
+function test_metrics() {
+    header "Checking metrics for total number of runs, should be 1 since we have spun up 1 run"
+    monitoring_token=`oc sa get-token prometheus-k8s -n openshift-monitoring`
```
anishasthana, line 57: I'm pretty sure this command will be breaking soon enough (openshift 4.11) @crobby we should consider an alternative command here right?

DharmitD: Where can I find documentation for OCP 4.11? Trying to look up for an alternative 4.11 supporting command but unable to find relevant info.

Also would the alternative 4.11 supporting command work for the earlier versions of OCP? Or will we just shift to 4.11 and not use earlier versions at all?

DharmitD: Do I need to replace ``` with `$()` around the command to make it OCP 4.11 compatible, like @crobby has done in his PR here? https://github.com/opendatahub-io/odh-manifests/blob/0bb13e5aa65f63e9b53b35a025ea3d2f746d0057/tests/basictests/modelmesh.sh#L65

anishasthana: Nah, you will probaably beed to manually grab the token from the secret associated with the service account. 

I am "fine" with us leaving it as is for now -- I think if tests break once we switch out ci over to 4.11 this is something we'd investigate.

anishasthana (commented):
/lgtm

tests/basictests/ml-pipelines.sh
crobby: Yeah, this should be fine.  I think it's deprecated functionality and we will replace all instances of this pattern with `oc create-token ...` in the future.

PR #642, opened 2022-08-15
Fix dashboard auth for service accounts

accorvin:
## Description
Previously, when using the `authentication` overlay for the dashboard, authn/z didn't work for OpenShift service accounts. I discovered that this was due to a misconfigured `--openshift-delegate-urls` value being passed to the dashboard. This change updates the argument to a valid value so that service accounts can be used with the dashboard.

## How Has This Been Tested?
This has been tested on an openshift 4.11 cluster by performing the following:

* Deploy the ODH operator to your cluster from the operator catalog
* Create a namespace for testing this change:
```
oc new-project auth-fix
```
* Create a kfdef to deploy the current version:
```
cat <<EOF | oc apply -f -
kind: KfDef
metadata:
  name: opendatahub
  namespace: auth-fix
spec:
  applications:
    - kustomizeConfig:
        overlays:
          - authentication
        repoRef:
          name: manifests
          path: odh-dashboard
      name: odh-dashboard
  repos:
    - name: manifests
      uri: 'https://github.com/opendatahub-io/odh-manifests/tarball/master'
EOF
```
* Run the following command to attempt to log into the dashboard using the `odh-dashboard` service account. The command should output `403`, indicating that the service account has been restricted from accessing the dashboard:
```
curl --header "Authorization: Bearer $(oc get secret -n auth-fix -o jsonpath='{.data.token}' `oc get secret -n auth-fix | grep odh-dashboard-token | awk '{print $1}'` | base64 -d)" -s -o /dev/null -w "%{http_code}\n" -k https://$(oc get route -n auth-fix odh-dashboard -o jsonpath='{.spec.host}')
```
* Update the kfdef to deploy this change by running the following:
```
cat <<EOF | oc apply -f -
kind: KfDef
metadata:
  name: opendatahub
  namespace: auth-fix
spec:
  applications:
    - kustomizeConfig:
        overlays:
          - authentication
        repoRef:
          name: manifests
          path: odh-dashboard
      name: odh-dashboard
  repos:
    - name: manifests
      uri: 'https://github.com/accorvin/idh-manifests/tarball/fix-dashboard-oauth'
EOF
```
* Run the curl command again. This time it should print a status code of `200`, indicating that the service account has successfully reached the dashboard:
```
curl --header "Authorization: Bearer $(oc get secret -n auth-fix -o jsonpath='{.data.token}' `oc get secret -n auth-fix | grep odh-dashboard-token | awk '{print $1}'` | base64 -d)" -s -o /dev/null -w "%{http_code}\n" -k https://$(oc get route -n auth-fix odh-dashboard -o jsonpath='{.spec.host}')
```
* Navigate to the dashboard in a browser. Confirm that you can log in and use the dashboard as normal.

## Merge criteria:


LaVLaS:
@accorvin Thx. Can you check this against the latest version of the ODH Dashboard [manifest](https://github.com/opendatahub-io/odh-dashboard/tree/main/manifests) and submit the PR there. We need to sync the most recent changes in odh-dashboard here.  

If this is a breaking change then we might need to backport the fix to the [v1.3 branch](https://github.com/opendatahub-io/odh-manifests/tree/v1.3)

accorvin:
@LaVLaS will do. Should I close this PR, or does the change need to be made to both places?

I don't think the change needs to be backported. Shouldn't break any existing flows, just fixes one that is currently broken.

accorvin:
Closing this a my understanding is that it should go into the odh-dashboard repo first and then there is a separate process for getting those changes into this repo. See https://github.com/opendatahub-io/odh-dashboard/pull/412

PR #649, opened 2022-08-16
Use default OLM channel for ocp pipelines subscription

LaVLaS:

## Description
Removes the explicit olm `channel` specification for the openshift-pipelines operator subscription to prevent any breaks due to the change or deprecation in any channel

## How Has This Been Tested?
Manual application of the subscription without the `channel:`

## Merge criteria:


anishasthana (commented):
/lgtm

shalberd:
@LaVLaS  removing the channel in general is an understandable idea. If I understand correctly, this means you at ODH fully support pipelines-1.7? 

"For example, the default channel for Red Hat OpenShift Pipelines Operator version 1.7 is pipelines-1.7"

https://docs.openshift.com/container-platform/4.11/cicd/pipelines/installing-pipelines.html

Elsewhere, you mentioned "The OpenShift Cluster must have OpenShift Pipelines 1.7.2 or higher installed."

@rimolive Maybe in that case, you should in the [readme](https://github.com/opendatahub-io/data-science-pipelines-operator/blob/main/README.md) reword to

"The OpenShift Cluster must have OpenShift Pipelines 1.7.2 or higher installed, within the 1.7 minor versions. We recommend 1.7.x versions, which is the default channel for the minor version of Red Hat Openshift Pipelines (pipelines-1.7)"

I know too little about backward-compatibility of e.g. pipelines-1.8 to pipelines-1.8, so having that mentioned explicitely might not be a bad idea.

PR #692, opened 2022-11-02
DashboardConfig CRD Changes for 2.4.0

andrewballantyne:
Manually sync the CRD changes from ODH Dashboard v2.4.0.

### DIFF
--- a/odh-dashboard/base/odh-dashboard-crd.yaml
+++ b/odh-dashboard/base/odh-dashboard-crd.yaml
@@ -41,6 +41,8 @@ spec:
                       type: boolean
                     disableUserManagement:
                       type: boolean
+                    disableProjects:
+                      type: boolean
                 groupsConfig:
                   type: object
                   required:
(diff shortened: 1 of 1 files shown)


DaoDaoNoCode (approved):
lgtm

anishasthana (commented):
/lgtm

anishasthana (approved):

LaVLaS:
@andrewballantyne @DaoDaoNoCode The next release is for ODH v1.4.1 to update dashboard to v2.3.1 and pull in the `*.dashboard.opendatahub.io` CRDs into the ODH "Owned" CRDs.  Is thi scompatible with that release?

DaoDaoNoCode:
I don't think we have DSG for dashboard 2.3.1 so the feature flag could probably not work anywhere because we don't have the feature at all, but it doesn't hurt to bring the CRD in IMO. Or you can also hold this until the release with dashboard 2.4.0

PR #708, opened 2022-11-29
Have etcd service select only etcd pods.

HumairAK:
Even though in their respective manifests etcd service selects on label `app: etcd` this is overwritten by this kustomization [setting](https://github.com/opendatahub-io/odh-manifests/blob/master/model-mesh/base/kustomization.yaml#L4). 

The result is the etcd service selects on all pods with the label `app: model-mesh`, this is a workaround change that will ensure etcd service only selects on etcd pods. A more robust solution would re organize the labels, and not add a broad `app: modelmesh` label on _all resources_ if the plan is to select on these labels by component.

## Merge criteria:

### DIFF
--- a/model-mesh/dependencies/quickstart.yaml
+++ b/model-mesh/dependencies/quickstart.yaml
@@ -15,6 +15,8 @@ apiVersion: v1
 kind: Service
 metadata:
   name: etcd
+  labels:
+    component: model-mesh-etcd
 spec:
   ports:
     - name: etcd-client-port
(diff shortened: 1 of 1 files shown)


LaVLaS:
/lgtm

CI tests will fail until I can resolve the dashboard CI test failures

HumairAK:
will send this change to the respective model-serving repositories instead to follow the regular flow of bringing component changes to upstream (i.e. sending it to the component repositories first)

HumairAK:
Re-opened as per discussion in mlops -- as this is a bug fix for already released content, this pr should be made now.

anishasthana (approved):
/lgtm

PR #720, opened 2023-01-23
Add kube-linter command to makefile

mlassak:
## Description
Added a manual command to makefile for kube-linter execution.

## Merge criteria:
- [x] The developer has manually tested the changes and verified that the changes work.


.kube-linter.yaml
```diff
@@ -0,0 +1,6 @@
+checks:
+  ignorePaths:
+    - .github/
+    - .aicoe-ci.yaml
```
anishasthana, line 4: Why do we have the same file twice?

In fact... it looks like we don't use it or the `.thoth.yaml` much at all. Should we remove all 3 of these from the repo? @LaVLaS

mlassak: I must have overlooked it when I was changing the config file at one point, I removed the duplicate file mention.

LaVLaS: Correct. `.aicoe-ci.yaml` and `.thoth.yaml` can be removed.  Can you add `tests/` to the list  also

Makefile
```diff
@@ -53,3 +56,8 @@ tag: update-kfdef-with-tag
 	git tag $(TAG_NAME)
 push-tag:
 	git push --tags
+
+# requires kube-linter to be installed (https://github.com/stackrox/kube-linter)
+lint:
```
LaVLaS, line 61: Can you change this target to `kube-lint`? 

```suggestion
kube-lint:
```

Makefile
```diff
@@ -9,10 +9,13 @@ REPO_URL="https://github.com/opendatahub-io/odh-manifests"
 KFDEF_FILE="kfdef/kfctl_openshift.yaml"
 COMMIT_MESSAGE="Update KFdef for release $(TAG_NAME)"
 UPDATE_TO_COMMIT=master
+KUBE_LINTER_CONFIG=.kube-linter.yaml
 ifndef VERSION
 $(help)
 endif
 
+.PHONY: lint
```
LaVLaS, line 17: Can you change this to `kube-lint`?

```suggestion
.PHONY: kube-lint
```

Makefile
```diff
@@ -53,3 +56,8 @@ tag: update-kfdef-with-tag
 	git tag $(TAG_NAME)
 push-tag:
 	git push --tags
+
+# requires kube-linter to be installed (https://github.com/stackrox/kube-linter)
+lint:
+	kube-linter lint ./ --config $(KUBE_LINTER_CONFIG)
```
LaVLaS, line 62: Can you add an additional optional argument that would allow us to pass arguments to `kube-linter`?  This would make it easier to target a specific directories instead of the default ALL directories

mlassak:
Thank you for your feedback, I changed the command names and added a new variable ```KUBE_LINTER_TARGET```, specification of directory/file to lint can be done using ```make kube-lint KUBE_LINTER_TARGET=path/to/target```.

PR #731, opened 2023-02-14
Use static tag of odh-notebook-controller manager image

LaVLaS:
## Description
Updates the odh-notebook-controller images to ensure that they are not using floating tags that could pull in updates that that are out of sync with the current manifest

## How Has This Been Tested?
* Deployed this locally to ensure existing functionality is preserved

## Merge criteria:
* ODH is deployed with notebook-controller and dashboard successfully
* User can spawn odh-core notebooks without any issues


odh-notebook-controller/odh-notebook-controller/manager/manager.yaml
```diff
@@ -17,7 +17,7 @@ spec:
         runAsNonRoot: true
       containers:
         - name: manager
-          image: quay.io/opendatahub/odh-notebook-controller:latest
+          image: quay.io/opendatahub/odh-notebook-controller:1.6-9ae9413
```
VaishnaviHire, line 20: Couple of questions:

- Should we be updating https://github.com/opendatahub-io/odh-manifests/blob/master/odh-notebook-controller/gen_odh_manifests.sh#L8 and then generate the manifests?
-  Do we need to update this for every PR build in controller, or before ODH release ?

LaVLaS: Good questions.  

1. For this PR, I just wanted to replace the `latest` tag for odh-notebook-controller but I should update that tag in gen_odh-manifests.sh for consistency.  
2. I'll see if running gen_odh-manifests.sh makes any drastic changes in the manifest.  If it does then I'll keep the update push this update to the next release of ODH and we can follow-up on  https://github.com/opendatahub-io/kubeflow/issues/56 to streamline the process of syncing the manifests from the notebook-controller repo to odh-manifests

PR #744, opened 2023-03-03
Data Science Pipelines auth-delegation crb fix

### LINKED ISSUE #743: Unable to authenticate to data science pipelines UI via bearer token
**Describe the bug**
When attempting to authenticate to the datacience pipelines UI from outside of the cluster utilizing a bearer token the request fails with a 403 error message.

**Steps To Reproduce**

1. Deploy cluster with ODH/data science pipelines
2. Login to the cluster via oc
3. Get user token with `oc whoami --show-token`
4. Use token to perform API call to server:

```
curl --location --request GET 'https://<cluster-url>/apis/v1beta1/experiments' \
--header 'Authorization: Bearer sha256~<token>' \
```

**Expected behavior**
Request should return the results of the API call.

**Workaround (if any)**
Any manual steps that allow you to resolve the issue

**Open Data Hub Version**
Please attach relevant kfdef manifest if applicable

**OpenShift Version**
Version #: 4.11
Provider (Baremetal, OpenStack, RHV, AWS, OKD, CodeReady Containers, ...): RHPDS

**Additional context**
When making the API call the oauth-proxy pod records the following error message:

```
E0303 22:25:43.173421 1 webhook.go:111] Failed to make webhook authenticator request: tokenreviews.authentication.k8s.io is forbidden: User "system:serviceaccount:mlops-demo-datascience:ds-pipeline-ui" cannot create resource "tokenreviews" in API group "authentication.k8s.io" at the cluster scope
2023/03/03 22:25:43 oauthproxy.go:793: requestauth: 10.128.2.16:55162 tokenreviews.authentication.k8s.io is forbidden: User "system:serviceaccount:mlops-demo-datascience:ds-pipeline-ui" cannot create resource "tokenreviews" in API group "authentication.k8s.io" at the cluster scope
```

It appears that the [auth-delegator](https://github.com/opendatahub-io/odh-manifests/blob/master/data-science-pipelines/overlays/ds-pipeline-ui/rolebindings/auth-delegator.yaml) crb contains the parameter `$(namespace)` which is not being replaced when deployed to openshift so the SA is not receiving the correct permissions.

strangiato:
## Description
Resolves #743

The auth-delegation for  CRB includes a parameter `$(namespace)` which is not being replaced by kustomize and is instead being rendered into the final yaml as `$(namespace)`.  This is causing the oauth pod to not have the correct permissions to allow users to authenticate via a bearer token from outside of the cluster.

This PR also fixes an issue where the CRB is not unique.  If multiple instances of ODH are deployed on the same cluster this non-unique name will cause the permissions to "swap" back and forth between the two different subject service accounts.

## How Has This Been Tested?
Deployed the manifest from my branch using the following KFDef:

```
kind: KfDef
metadata:
  name: opendatahub
spec:
  applications:
    - kustomizeConfig:
        repoRef:
          name: manifests
          path: odh-common
      name: odh-common
    - kustomizeConfig:
        repoRef:
          name: manifests
          path: odh-dashboard
      name: odh-dashboard
    - kustomizeConfig:
        repoRef:
          name: manifests
          path: odh-notebook-controller
      name: odh-notebook-controller
    - kustomizeConfig:
        repoRef:
          name: manifests
          path: notebook-images
      name: notebook-images
    - kustomizeConfig:
        overlays:
          - odh-model-controller
        repoRef:
          name: manifests
          path: model-mesh
      name: model-mesh
    - kustomizeConfig:
        overlays:
          - metadata-store-mariadb
          - ds-pipeline-ui
          - object-store-minio
          - integration-odhdashboard
          - default-configs
        repoRef:
          name: manifests
          path: data-science-pipelines
      name: data-science-pipelines
  repos:
    - name: manifests
      uri: 'https://github.com/strangiato/odh-manifests/tarball/oauth-crb-fix'
```

Validated CRB by manually inspecting CRB after it had been deployed to OpenShift and performed an API call as detailed in #743

## Merge criteria:


LaVLaS:
@strangiato We are deprecating this individual deployment of data-science-pipelines in favor of a dedicated data-science-pipelines operator managing multiple deployments of data-science-pipelines in individual namespaces

CC @rimolive @HumairAK Do you need to factor this change into the operator deploying data-science-pipelines

rimolive:
@LaVLaS @strangiato The old kfp UI will soon be replaced by the `odh-dashboard`, which will make everything integrated in a single UI.

As for the change, if it's caused by lack of permissions to access the UI, I could swear it was fixed before. If it's a CLI issue, then I'm ok to merge as it might be something to consider when integrating with the new UI.

strangiato:
@rimolive I am attempting to access the KFP api using the python kfp package.  

Since the UI is utilizing the oauth-proxy, you have to authenticate with a bearer token from OpenShift (e.g. `oc whoami --show-token`).  The oauth-proxy was configured to allow a user to authenticate, but it was throwing an error message related to some missing permissions in the oauth pod.  There is a CRB that was intended to grant the correct permissions but it was missing some additional configuration for `namespace: $(namespace)`.  The `$(namespace)` was not getting replaced with the namespace name and instead `$(namespace)` is getting rendered into the final YAML on OCP as a literal string.

In case you are curious what the code looks like for what I was trying to do, here is was I was putting together:
https://github.com/rh-intelligent-application-practice/kubeflow-examples/blob/main/pipelines/test_connection.py

The above code does work correctly with by fix btw.

As far as replacing the KFP UI with odh-dashboard, I assume that the `/pipeline` API will still be exposed somewhere, correct?

HumairAK:
Just tried building the kfdef upstream, indeed it doesn't consupe the parameters, I think this change is good to keep the deprectaed dsp in working order until we officially switch to the operator, thanks @strangiato ! 

These changes are not required for the operator provisioned dsp, as both uniqueness and namespace for the SA are successfully substituted and accounted for. but as the DSPO is very new, we definitely [welcome others to try it out and confirm for themselves](https://github.com/opendatahub-io/odh-manifests/tree/master/data-science-pipelines-operator#deploy-the-operator-via-odh), @strangiato. 


/lgtm
/approve

strangiato:
Any suggestions for resolving the failed tests?

It looks like #749 is failing with the same errors so I assume this isn't related to my changes...

LaVLaS:
@strangiato This requires a rebase since it is missing the new dsp-operator component

strangiato:
@LaVLaS Thanks!  Rebased from master.

LaVLaS:
/retest

CI infra issues caused the failure

HumairAK:
Looks like the pipelines tests (concerning this pr) are passing but operator tests (unrelated to this pr) failed, looking into this

anishasthana (approved):

Issue #745, opened 2023-03-08
Add a second tag on base ImageStreams to support a second version of the image


atheo89:
Update ImageStreams objects for the core images to have a second tag in order to support one more updated version of the image

shalberd:
if you do this the way you currently do it in the pull request, 
you potentially make airgapped / disconnected installs that rely on unique and unchanging image references via SHA256 digest more complicated or even impossible again ... tag.from.name should reference in digest.format.
If anything, when you add a second actual, more dynamic weekly imagestream tag that then points to the new sha256 image value.
https://github.com/opendatahub-io/odh-manifests/blob/4aa21564b0f23b413b63e5766677b6f6e39c6af9/notebook-images/base/jupyter-datascience-notebook-imagestream.yaml#L24

Name the imagestream tag-value to equal the image-tag value in the from.name.
e.g. from
```
from:
      kind: DockerImage
      name: quay.io/opendatahub/workbench-images:jupyter-datascience-ubi9-python-3.9-2023a-weekly
    name: "py3.9-v2"
```
to
```
from:
      kind: DockerImage
      name: quay.io/opendatahub/workbench-images@sha256:655f961295409d0d3a14927123ead8fff1b50b966948bcf18d10909e3ab32954
    name: "jupyter-datascience-ubi9-python-3.9-2023a-weekly"
```
![Bildschirmfoto 2023-03-24 um 18 27 28](https://user-images.githubusercontent.com/21118431/227597923-dda2848c-0868-4fce-88d7-df030818d964.png)

That way, you have visibility as to what the actual tag is and only need to ensure you update the digest-value in from.name as the tag changes from week to week.

Plus, we have ImageContentSourcePolicy and airgapped / disconnected support.

I have talked about this issue in another context with @VaishnaviHire 

`[Background](https://bugzilla.redhat.com/show_bug.cgi?id=1741391)`:

Description of problem:
In disconnected environments, CRIO uses the imagecontentsourcepolicy to "fallback" to alternate registry sources when pulling sha-based image specs.

The imagestream import controller should do the same, again only when importing a SHA based imagestreamtag.  If the external references is a SHA, and the import against the original registry fails, it should use the imagecontentsourcepolicy to try alternate registry sources to import the SHA.

If it succeeds, then the "status" for the tag should reflect the alternate registry.

@Jooho Can you tell us your opinion, please? See the suggested changes in the linked PR.

shalberd:
[...]

PR #746, opened 2023-03-08
Add a second tag on OOTB images to support a second version N & N-1

atheo89:
Issue: https://github.com/opendatahub-io/odh-manifests/issues/745

## Description
This PR introduces a second recommended version of the image. We fetch the image using the tag as it aligns with having faster ODH releases.

For example: 
`name: quay.io/opendatahub/workbench-images:jupyter-datascience-ubi9-python-3.9-2023a-weekly`

## How Has This Been Tested?

   1. Deploy ODH with enabled the notebook-images
   2. Login to odh-dashboard and spawn an OOTB image ensure that there is a second version 
   3. Confirm that the notebook spawned successfully and OOTB Image functionality works

## Merge criteria:

### DIFF
--- a/notebook-images/base/jupyter-datascience-notebook-imagestream.yaml
+++ b/notebook-images/base/jupyter-datascience-notebook-imagestream.yaml
@@ -4,7 +4,7 @@ metadata:
   labels:
     opendatahub.io/notebook-image: "true"
   annotations:
-    opendatahub.io/notebook-image-url: "https://github.com//opendatahub-io/notebooks/tree/main/jupyter/datascience/ubi8-python-3.8"
+    opendatahub.io/notebook-image-url: "https://github.com//opendatahub-io/notebooks/tree/main/jupyter/datascience"
     opendatahub.io/notebook-image-name: "Jupyter Data Science"
     opendatahub.io/notebook-image-desc: "Jupyter notebook image with a set of data science libraries that advanced AI/ML notebooks will use as a base image to provide a standard for libraries avialable in all notebooks"
     opendatahub.io/notebook-image-order: "2"
@@ -13,6 +13,19 @@ spec:
   lookupPolicy:
     local: true
   tags:
+  # N Version of the image
+  - annotations:
+      opendatahub.io/notebook-software: '[{"name":"Python","version":"v3.9"}]'
+      opendatahub.io/notebook-python-dependencies: '[{"name":"Boto3","version":"1.26"},{"name":"Kafka-Python","version":"2.0"},{"name":"Matplotlib","version":"3.6"},{"name":"Numpy","version":"1.24"},{"name":"Pandas","version":"1.5"},{"name":"Scikit-learn","version":"1.2"},{"name":"Scipy","version":"1.10"}]'
+      openshift.io/imported-from: quay.io/opendatahub/workbench-images
+      opendatahub.io/workbench-image-recommended: 'true'
+    from:
+      kind: DockerImage
+      name: quay.io/opendatahub/workbench-images:jupyter-datascience-ubi9-python-3.9-2023a-weekly
+    name: "py3.9-v2"
+    referencePolicy:
+      type: Local
+  # N-1 Version of the image
   - annotations:
       opendatahub.io/notebook-software: '[{"name":"Python","version":"v3.8"}]'
       opendatahub.io/notebook-python-dependencies: '[{"name":"Boto3","version":"1.17"},{"name":"Kafka-Python","version":"2.0"},{"name":"Matplotlib","version":"3.4"},{"name":"Numpy","version":"1.19"},{"name":"Pandas","version":"1.2"},{"name":"Scikit-learn","version":"0.24"},{"name":"Scipy","version":"1.6"}]'
@@ -22,4 +35,4 @@ spec:
       name: quay.io/opendatahub/notebooks@sha256:5df71f5542d2e0161f0f4342aa9a390679d72dc6fae192fd8da1e5671b27e8d4
     name: "py3.8-v1"
     referencePolicy:
-      type: Source
+      type: Local
--- a/notebook-images/base/jupyter-minimal-notebook-imagestream.yaml
+++ b/notebook-images/base/jupyter-minimal-notebook-imagestream.yaml
@@ -4,7 +4,7 @@ metadata:
   labels:
     opendatahub.io/notebook-image: "true"
   annotations:
-    opendatahub.io/notebook-image-url: "https://github.com//opendatahub-io/notebooks/tree/main/jupyter/minimal/ubi8-python-3.8"
+    opendatahub.io/notebook-image-url: "https://github.com//opendatahub-io/notebooks/tree/main/jupyter/minimal"
     opendatahub.io/notebook-image-name: "Minimal Python"
     opendatahub.io/notebook-image-desc: "Jupyter notebook image with minimal dependency set to start experimenting with Jupyter environment."
     opendatahub.io/notebook-image-order: "1"
@@ -13,11 +13,24 @@ spec:
   lookupPolicy:
     local: true
   tags:
+  # N Version of the image
+  - annotations:
+      opendatahub.io/notebook-software: '[{"name":"Python","version":"v3.9"}]'
+      opendatahub.io/notebook-python-dependencies: '[{"name":"JupyterLab","version": "3.5"}, {"name": "Notebook","version": "6.5"}]'
+      openshift.io/imported-from: quay.io/opendatahub/workbench-images
+      opendatahub.io/workbench-image-recommended: 'true'
+      opendatahub.io/default-image: "true"
+    from:
+      kind: DockerImage
+      name: quay.io/opendatahub/workbench-images:jupyter-minimal-ubi9-python-3.9-2023a-weekly
+    name: "py3.9-v2"
+    referencePolicy:
+      type: Local
+  # N-1 Version of the image
   - annotations:
       opendatahub.io/notebook-software: '[{"name":"Python","version":"v3.8"}]'
       opendatahub.io/notebook-python-dependencies: '[{"name":"JupyterLab","version": "3.2"}, {"name": "Notebook","version": "6.4"}]'
       openshift.io/imported-from: quay.io/opendatahub/notebooks
-      opendatahub.io/default-image: "true"
     from:
       kind: DockerImage
       name: quay.io/opendatahub/notebooks@sha256:a138838e1c9acd7708462e420bf939e03296b97e9cf6c0aa0fd9a5d20361ab75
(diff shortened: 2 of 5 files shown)


notebook-images/base/jupyter-pytorch-notebook-imagestream.yaml
```diff
@@ -13,6 +13,19 @@ spec:
...
+  # N Version of the image
+  - annotations:
+      opendatahub.io/notebook-software: '[{"name":"Python","version":"v3.9"},{"name":"PyTorch","version":"1.13"}]'
+      opendatahub.io/notebook-python-dependencies: '[{"name":"PyTorch","version":"1.13"},{"name":"Tensorboard","version":"2.11"},{"name":"Boto3","version":"1.26"},{"name":"Kafka-Python","version":"2.0"},{"name":"Matplotlib","version":"3.6"},{"name":"Numpy","version":"1.24"},{"name":"Pandas","version":"1.5"},{"name":"Scikit-learn","version":"1.2"},{"name":"Scipy","version":"1.10"}]'
+      openshift.io/imported-from: quay.io/opendatahub/workbench-images
+      opendatahub.io/workbench-image-recommended: 'true'
+    from:
+      kind: DockerImage
+      name: quay.io/opendatahub/workbench-images:jupyter-pytorch-ubi9-python-3.9-2023a-weekly
+    name: "py3.9-v1"
```
VaishnaviHire, line 25: Should this be tagged as `py3.9-v2` ?

atheo89: Done

notebook-images/base/jupyter-minimal-notebook-imagestream.yaml
```diff
@@ -13,6 +13,20 @@ spec:
...
+    from:
+      kind: DockerImage
+      name: quay.io/opendatahub/workbench-images:jupyter-minimal-ubi9-python-3.9-2023a-weekly
+    name: "py3.9-v2"
+    referencePolicy:
+      type: Local
+  # N-1 Version of the image
   - annotations:
       opendatahub.io/notebook-software: '[{"name":"Python","version":"v3.8"}]'
       opendatahub.io/notebook-python-dependencies: '[{"name":"JupyterLab","version": "3.2"}, {"name": "Notebook","version": "6.4"}]'
```
VaishnaviHire, line 32: ` opendatahub.io/default-image: "true"` annotation needs to be removed from n-1, else it sets the n-1 version as default

atheo89: Done

VaishnaviHire:
Looks good. I think the branch needs a re-base

notebook-images/base/jupyter-datascience-notebook-imagestream.yaml
```diff
@@ -13,6 +13,19 @@ spec:
...
+  - annotations:
+      opendatahub.io/notebook-software: '[{"name":"Python","version":"v3.9"}]'
+      opendatahub.io/notebook-python-dependencies: '[{"name":"Boto3","version":"1.26"},{"name":"Kafka-Python","version":"2.0"},{"name":"Matplotlib","version":"3.6"},{"name":"Numpy","version":"1.24"},{"name":"Pandas","version":"1.5"},{"name":"Scikit-learn","version":"1.2"},{"name":"Scipy","version":"1.10"}]'
+      openshift.io/imported-from: quay.io/opendatahub/workbench-images
+      opendatahub.io/workbench-image-recommended: 'true'
+    from:
+      kind: DockerImage
+      name: quay.io/opendatahub/workbench-images:jupyter-datascience-ubi9-python-3.9-2023a-weekly
+    name: "py3.9-v2"
+    referencePolicy:
```
shalberd, line 26: referencePolicy should be set to type: Local, otherwise, the image is always pulled-through from external

shalberd:
if you do this, you potentially make airgapped / disconnected installs that rely on unique and unchanging image references via SHA256 digest more complicated or even impossible again ... tag.from.name should reference in digest.format.
If anything, really add a second actual imagestream tag that then points to the new sha256 image value.

`[Background](https://bugzilla.redhat.com/show_bug.cgi?id=1741391)`:

Description of problem:
In disconnected environments, CRIO uses the imagecontentsourcepolicy to "fallback" to alternate registry sources when pulling sha-based image specs.

The imagestream import controller should do the same, again only when importing a SHA based imagestreamtag.  If the external references is a SHA, and the import against the original registry fails, it should use the imagecontentsourcepolicy to try alternate registry sources to import the SHA.

If it succeeds, then the "status" for the tag should reflect the alternate registry.

@Jooho Can you tell us your opinion, please? See the suggested changes.

notebook-images/base/jupyter-datascience-notebook-imagestream.yaml
```diff
@@ -13,6 +13,19 @@ spec:
...
   tags:
+  # N Version of the image
+  - annotations:
+      opendatahub.io/notebook-software: '[{"name":"Python","version":"v3.9"}]'
+      opendatahub.io/notebook-python-dependencies: '[{"name":"Boto3","version":"1.26"},{"name":"Kafka-Python","version":"2.0"},{"name":"Matplotlib","version":"3.6"},{"name":"Numpy","version":"1.24"},{"name":"Pandas","version":"1.5"},{"name":"Scikit-learn","version":"1.2"},{"name":"Scipy","version":"1.10"}]'
+      openshift.io/imported-from: quay.io/opendatahub/workbench-images
+      opendatahub.io/workbench-image-recommended: 'true'
+    from:
+      kind: DockerImage
+      name: quay.io/opendatahub/workbench-images:jupyter-datascience-ubi9-python-3.9-2023a-weekly
```
shalberd, line 24: this should always be in digest format @sha256 to support airgapped and ImageContentSourcePolicy under OpenShift.
Refer to the current digest value here in tag.from.name  `quay.io/opendatahub/workbench-images@sha256:655f961295409d0d3a14927123ead8fff1b50b966948bcf18d10909e3ab32954`

notebook-images/base/jupyter-datascience-notebook-imagestream.yaml
```diff
@@ -13,6 +13,19 @@ spec:
...
+    name: "py3.9-v2"
```
shalberd, line 25: use the image-tag as the imagestream tag
`name: jupyter-datascience-ubi9-python-3.9-2023a-weekly`

In my opinion, such weekly tags are bad practice in deployment references. But, you could implement a process to get the sha256 digest value each week and then update tag.from.name. I do see why tags get updated to link to new images in CICD, just do not use them in tag.from.name

LaVLaS:
@shalberd I just want to reiterate that we completely understand your concerns/requirements for using image digest hashes for all images to support disconnected|airgapped environments.  On the odh-manifests `master` branch we will not be using image digest hashes for the notebook image uri since they will be rebuilt daily/weekly to address any security issues.  You can expect that we will be converting to image digest hashes after creating an odh-manifest release
to ensure stability in all of the images deploy in a stable ODH release.

notebook-images/base/jupyter-datascience-notebook-imagestream.yaml
LaVLaS: See https://github.com/opendatahub-io/odh-manifests/pull/746#issuecomment-1484430943 about using digest image hashes odh-manifests releases

notebook-images/base/jupyter-datascience-notebook-imagestream.yaml
LaVLaS: See https://github.com/opendatahub-io/odh-manifests/pull/746#issuecomment-1484430943 about using digest image hashes odh-manifests releases

shalberd:
GUI support for multiple tags per imagestream in there now, nice. 

https://github.com/opendatahub-io/odh-dashboard/pull/992#issuecomment-1484486678

Also, since you keep this to the manifests master branch only, it is fine.

Thank you, @LaVLaS for the clarification.

Have a good start of the week.

notebook-images/base/jupyter-datascience-notebook-imagestream.yaml
shalberd: Totally fine and understood. The imagestream (not image reference) tag itself, though, should be informative as to the image-tag referenced. It all depends on what you want to achieve with upstream RHOSDS check of env variable JUPYTER_IMAGE.

notebook-images/base/jupyter-tensorflow-notebook-imagestream.yaml
```diff
@@ -13,10 +13,23 @@ spec:
...
-        opendatahub.io/notebook-python-dependencies: '[{"name":"TensorFlow","version":"2.7"},{"name":"Tensorboard","version":"2.6"},{"name":"Boto3","version":"1.17"},{"name":"Kafka-Python","version":"2.0"},{"name":"Matplotlib","version":"3.4"},{"name":"Numpy","version":"1.19"},{"name":"Pandas","version":"1.2"},{"name":"Scikit-learn","version":"0.24"},{"name":"Scipy","version":"1.6"}]'
-        openshift.io/imported-from: quay.io/opendatahub/notebooks
+      opendatahub.io/notebook-software: '[{"name":"Python","version":"v3.9"},{"name":"TensorFlow","version":"2.11"}]'
+      opendatahub.io/notebook-python-dependencies: '[{"name":"TensorFlow","version":"2.11"},{"name":"Tensorboard","version":"2.11"},{"name":"Boto3","version":"1.26"},{"name":"Kafka-Python","version":"2.0"},{"name":"Matplotlib","version":"3.6"},{"name":"Numpy","version":"1.24"},{"name":"Pandas","version":"1.5"},{"name":"Scikit-learn","version":"1.2"},{"name":"Scipy","version":"1.10"}]'
+      openshift.io/imported-from: quay.io/opendatahub/workbench-images
+      opendatahub.io/workbench-image-recommended: 'true'
```
shalberd, line 21: this and its representation in ODH Dashboard GUI is very nice, good idea.

notebook-images/base/jupyter-tensorflow-notebook-imagestream.yaml
```diff
@@ -13,10 +13,23 @@ spec:
...
+    from:
+      kind: DockerImage
+      name: quay.io/opendatahub/workbench-images:cuda-jupyter-tensorflow-ubi9-python-3.9-2023a-weekly
+    name: "py3.9-v2"
```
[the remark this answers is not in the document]
LaVLaS: I think the current intent is that Imagestream name will be the global identifier for the major package in the notebook (`Tensorflow`) and the multiple tag names will reference the difference in the base images (python 3.8 vs python 3.9)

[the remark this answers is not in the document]
shalberd: I see, and yes, image tag and imagestream tag do not have to equal. For cases outside continuous development, in the stable branches, when you have digest-format in tag.from.name, you could maybe put a yaml comment above it with the image-tag.

LaVLaS:
/lgtm

Readding @VaishnaviHire  /lgtm after the rebase

PR #748, opened 2023-03-10
Fix ds-pipeline-viewer-crd-service-account authorization to add final…

adrien-legros:
…izer on owned deployments

Allow ds-pipeline-viewer-crd-service-account to add finalizers on the deployments it creates.

## Description
With current role and rolebinding, the service account cannot add a finalizer on their deployments. Therefore the deployment fail and we have an error on the ds-pipeline-viewer-crd: "cannot set blockOwnerDeletion if an ownerReference refers to a resource you can't set finalizers on".
Also, I found out that upstream Kubeflow Pipeline also add this resource to the role: https://github.com/kubeflow/pipelines/blob/master/manifests/kustomize/base/pipeline/ml-pipeline-viewer-crd-role.yaml.

## How Has This Been Tested?
Tested manually only. But upstream have same settings Kubeflow Pipeline

### DIFF
--- a/data-science-pipelines/overlays/component-visualization-server/roles/ds-pipeline-viewer-controller.yaml
+++ b/data-science-pipelines/overlays/component-visualization-server/roles/ds-pipeline-viewer-controller.yaml
@@ -22,6 +22,7 @@ rules:
       - kubeflow.org
     resources:
       - viewers
+      - viewers/finalizers
     verbs:
       - create
       - get


HumairAK:
/lgtm

@rimolive

rimolive:
Seems like it needs a rebase, but the fix looks food to me.

LaVLaS:
I moved this to target the [v1.5 branch](https://github.com/opendatahub-io/odh-manifests/tree/v1.5) since this manifest will be removed after the ODH v1.5 release

LaVLaS:
Closing this PR since the data-science-pipelines manifest was deprecated in the ODH 1.5.0 release and replaced with the [data-science-pipelines-operator](https://github.com/opendatahub-io/data-science-pipelines-operator)
@adrien-legros If you have any issues with the new namespaced deployment(s) of data-science-pipelines feel free to file an issue under [data-science-pipelines-operator](https://github.com/opendatahub-io/data-science-pipelines-operator/issues)

PR #768, opened 2023-03-30
Change fallback OVMS URL to point to our built image

Xaenalt:
Image includes Nvidia GPU inferencing support

### DIFF
--- a/model-mesh/base/params.env
+++ b/model-mesh/base/params.env
@@ -3,6 +3,6 @@ monitoring-namespace=
 odh-mm-rest-proxy=quay.io/opendatahub/rest-proxy:v0.11.0-alpha
 odh-modelmesh-runtime-adapter=quay.io/opendatahub/modelmesh-runtime-adapter:v0.11.0-alpha
 odh-modelmesh=quay.io/opendatahub/modelmesh:v0.11.0-alpha
-odh-openvino=quay.io/opendatahub/openvino_model_server:2022.3-gpu
+odh-openvino=quay.io/opendatahub/openvino_model_server@sha256:20dbfbaf53d1afbd47c612d953984238cb0e207972ed544a5ea662c2404f276d
 odh-modelmesh-controller=quay.io/opendatahub/modelmesh-controller:v0.11.0-alpha
 odh-model-controller=quay.io/opendatahub/odh-model-controller:v0.9.6-auth
--- a/model-mesh/runtimes/kustomization.yaml
+++ b/model-mesh/runtimes/kustomization.yaml
@@ -26,8 +26,8 @@ images:
     newTag: 0.5.2
 
   - name: ovms-1
-    newName: openvino/model_server
-    newTag: "2022.3"
+    newName: quay.io/opendatahub/openvino_model_server
+    newTag: "2022.3-release"
 
 transformers:
   - ../default/metadataLabelTransformer.yaml


Xaenalt:
I'm also not completely convinced we need this per se too, we may want the fallback image to point elsewhere, but I'll let the reviewers decide on that

Jooho (approved):
/lgtm

LaVLaS:
@Xaenalt We need to upstream this repo and host these images in quay.io/opendatahub instead of pulling from a downstream image repository for a build system that is already in use upstream

model-mesh/base/params.env
```diff
@@ -3,6 +3,6 @@ monitoring-namespace=
 odh-mm-rest-proxy=quay.io/opendatahub/rest-proxy:v0.11.0-alpha
 odh-modelmesh-runtime-adapter=quay.io/opendatahub/modelmesh-runtime-adapter:v0.11.0-alpha
 odh-modelmesh=quay.io/opendatahub/modelmesh:v0.11.0-alpha
-odh-openvino=quay.io/opendatahub/openvino_model_server:2022.3-gpu
+odh-openvino=quay.io/modh/openvino-model-server@sha256:c89f76386bc8b59f0748cf173868e5beef21ac7d2f78dada69089c4d37c44116
```
Jooho, line 6: I pushed the image into opendatahub.
Use this quay.io/opendatahub/openvino_model_server@sha256:20dbfbaf53d1afbd47c612d953984238cb0e207972ed544a5ea662c2404f276d

model-mesh/runtimes/kustomization.yaml
```diff
@@ -26,8 +26,8 @@ images:
     newTag: 0.5.2
 
   - name: ovms-1
-    newName: openvino/model_server
-    newTag: "2022.3"
+    newName: quay.io/modh/openvino-model-server
```
Jooho, line 29: quay.io/opendatahub/openvino_model_server

Xaenalt:
Let me rebase it onto main, that might help

LaVLaS:
I didn't investigate the root cause but it looks like the operator was not installed successfully OR the odh-core `kfdef` was never deployed so there are no components

LaVLaS:
@Xaenalt A rebase should kick off a new test run

model-mesh/base/params.env
```diff
@@ -3,6 +3,6 @@ monitoring-namespace=
 odh-mm-rest-proxy=quay.io/opendatahub/rest-proxy:v0.11.0-alpha
 odh-modelmesh-runtime-adapter=quay.io/opendatahub/modelmesh-runtime-adapter:v0.11.0-alpha
 odh-modelmesh=quay.io/opendatahub/modelmesh:v0.11.0-alpha
-odh-openvino=quay.io/opendatahub/openvino_model_server:2022.3-gpu
+odh-openvino=quay.io/opendatahub/openvino_model_server@sha256:20dbfbaf53d1afbd47c612d953984238cb0e207972ed544a5ea662c2404f276d
```
shalberd, line 6: why is there sometime tag-notation and sometimes @sha256 digest notation in the manifests? I don't see a clear line of thought here.

model-mesh/runtimes/kustomization.yaml
```diff
@@ -26,8 +26,8 @@ images:
     newTag: 0.5.2
 
   - name: ovms-1
-    newName: openvino/model_server
-    newTag: "2022.3"
+    newName: quay.io/opendatahub/openvino_model_server
+    newTag: "2022.3-release"
```
shalberd, line 30: in any non-master branch, digest: sha256xxxxxxxx format instead of newTag format should be used to support restricted network environment installs (on-prem) with mirrored repos and ImageContentSourcePolicy.

model-mesh/base/params.env
Jooho: params.env file is for optional input from kfdef file. Openvino is not changed frequently so it is ok to use sha256 directly. however, for the others, we are thinking to use the doc to support a disconnected environment. In other word,  the image information in this file is just for default value but we can change it with sha256 with kfdef to support a disconnected environment.

model-mesh/runtimes/kustomization.yaml
Jooho: this information has nothing to do with a disconnected environment.  The mirror command line is not referring this information.

shalberd: but the **ImageContentSourcePolicy generated by oc mirror**, or, in my case generated manually, needs sha256 type digests, not tags, in manifests. Note my content "restricted network environment installs (on-prem) with mirrored repos and ImageContentSourcePolicy"

model-mesh/runtimes/kustomization.yaml
```diff
@@ -26,8 +26,8 @@ images:
     newTag: 0.5.2
 
   - name: ovms-1
```
shalberd, line 28: that name is nowhere in the image field of https://github.com/opendatahub-io/odh-manifests/blob/master/model-mesh/runtimes/ovms-1.x.yaml#L37 instead, there is a kustomize parameter value in there from https://github.com/opendatahub-io/odh-manifests/blob/master/model-mesh/base/params.env#L6 Same goes for the other three ServingRunetime yamls referenced ...

PR #772, opened 2023-04-04
Add GitHub action to update notebook's ImagesStream tag name field from tag into digest format

atheo89:
Add GitHub action to update notebook's ImagesStream tag name field from tag into digest format

For example:
**From Main** branch: `name: quay.io/opendatahub/workbench-images:jupyter-minimal-ubi9-python-3.9-2023a-weekly` 
**To Release** branch:  `name: "quay.io/opendatahub/workbench-images@sha256:46e2e9ac6b6b849c4dfafc47f03e3654895f9d61138322d677c559dae741fb26"`
## Description
The aim of this GitHub workflow is to update the .spec.tags[0].from.name field of each Notebook's ImageStreams into Digest format, when cut off a new release (the odh-manifest release branch should exist). 

The format of the digest that would be assing has the following format: quay.io/opendatahub/workbench-images@sha256:xxx...xxx

When it is finished it opens a Pull request. 

Note: Please note the older version of the images, which is n-1, belongs to the old implementation and cannot support the weekly updates. However, we plan to have an n-1 configuration available when the current N version becomes outdated. Therefore, the 2023a release will be based on the n-1 configuration, while the 2023b release will be based on the new N configuration. Eventually, we can apply a new job named: `update-n-1-version` to update the `.spec.tags[1].from.name' as well.

## How Has This Been Tested?
1. The user should manually trigger the Action, and provide as input the release branch that wants to update the tag names.
 
image

2. Ensure that all jobs passed:

image

3. Evaluate the new commits on the 'release'

image

## Merge criteria:


.github/workflows/digest-updater.yaml
```diff
@@ -0,0 +1,87 @@
...
+          #Fetch the latest tag in Digest format: ex.digest=quay.io/opendatahub/workbench-images@sha256:xxx...xxx
+          digest=$(skopeo inspect docker://quay.io/opendatahub/workbench-images:$latest_tag | jq .Digest | sed 's/sha256/quay.io\/opendatahub\/workbench-images@sha256/g')
+          digest=$(echo "$digest" | tr -d '"' | tr -d '\\')
+          echo $digest
+
+          # Update the .spec.tags[0].from.name field on the release image stream with the digest outputed digest
+          yq eval '.spec.tags[0].from.name = "'$digest'"' -i ${{ matrix.files }}
+
+          # Fetch, pull, add, and commit the changes into the release branch
+          git fetch origin ${{ inputs.branch }} && git pull origin ${{ inputs.branch}} && git add ${{ matrix.files }} && git commit -m "Update file ${{ matrix.files }} via digest-updater GitHub action" && git push origin ${{ inputs.branch }}
```
LaVLaS, line 86: Can we change this a create a Pull Request instead of merging directly to the release branch as a final check to make sure that there is no error in this action

.github/workflows/digest-updater.yaml
```diff
@@ -0,0 +1,87 @@
...
+
+      - name: Fetch digest, and update the imagestream file (Step 2)
+        env:
+          IMG_NAME: ${{ steps.img_name.outputs.img_name }}
+        run: |
+          src_tag=$(echo "${{ env.IMG_NAME }}" | sed 's/-weekly.*//' | cut -d':' -f2 | tr -d '"')
+          regex="$src_tag-\d+-${{ steps.hash.outputs.HASH }}"
+          echo $regex
+
+          # Get the latest tag: output ex. latest_tag=jupyter-minimal-ubi9-python-3.9-2023a-20230403-cf92364
```
LaVLaS, line 73: If we are using `*-weekly` tags for the imagestreams in odh-manifests and with the expectation that it is a floating tag that will be updated weekly, can we simplify this to  use the `skopeo inspect <IMAGE:TAG> | jq -r .Digest` as the latest digest hash?

atheo89: The weekly isn't a floating tag, its lifetime is only for one week then replaces by the new rebuild, thus the digest doesn't remain stable over the time. The `notebook-release-YYYYMMDD-git-hash` on the other side remains the same always.

.github/workflows/digest-updater.yaml
atheo89: Right, I updated the workflow to open a pull request instead. 
image

.github/workflows/digest-updater.yaml
shalberd: in my view, floating tag means exactly that: the the tag itself is fixed, e.g. blabla-weekly, but the digest behind the image changes regularly. See https://redhat-connect.gitbook.io/catalog-help/container-images/container-image-details/image-tags-and-versions. Very good stat release-type images stay unchanged and do not chnage their sha digest value.

shalberd:
What do you mean by a "release" branch? v1.5, for example? Have you given thought not only to add digest-values, but to also insert the actual image-tag from the remote repository as a yaml comment above from.name?

atheo89:
Hi @shalberd, thanks for the review! The release branch is an example name, in reality, will be the v1.5. About the comment, we don't need it into yaml because if you copy the sha and paste it into the quay search bar it will show which image is with the full name.

atheo89:
I am closing this PR because accidentally I pushed in the wrong remote... :cold_sweat: I will reopen it soon.

PR #784, opened 2023-04-11
Add data-science-pipelines-operator label for DSPO manifests

DharmitD:
## Description

If we do an `oc get pods -l app.kubernetes.io/created-by=data-science-pipelines-operator` after deploying DSP Operator, it won't return any pods, because only the DSPO deployment contains this label, the pods don't. Adding this label to the pods as well, so they could be filtered using the component name, that is, Data Science Pipelines Operator.

## How Has This Been Tested?

- Deploy a DSP Operator instance.
- Try out `oc get pods -l app.kubernetes.io/created-by=data-science-pipelines-operator` in the same namespace where the operator is deployed. Make sure it displays DSPO pods.

## Merge criteria:

### DIFF
--- a/data-science-pipelines-operator/manager/manager-service.yaml
+++ b/data-science-pipelines-operator/manager/manager-service.yaml
@@ -4,6 +4,7 @@ metadata:
   name: data-science-pipelines-operator
   labels:
     control-plane: controller-manager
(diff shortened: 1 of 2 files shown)


harshad16 (approved):
/lgtm 
thanks :+1:

PR #785, opened 2023-04-12
Fix tag's names on notebooks imagestreams with better naming

atheo89:
This PR fixes the tag names on notebook versions, for better presentation on the notebooks server UI.

## Description
For example, instead of being `py3.8-v2` which doesn't provide any useful information changed to `2023.1` which is the version of the notebook (the N version for example) 

## How Has This Been Tested?

-  Deploy ODH with enabled the notebook-images
-  Login to odh-dashboard and navigate into notebooks server spawner
-  Check the version name that has been changed 

image

 
## Merge criteria:

### DIFF
--- a/notebook-images/base/jupyter-datascience-notebook-imagestream.yaml
+++ b/notebook-images/base/jupyter-datascience-notebook-imagestream.yaml
@@ -22,7 +22,7 @@ spec:
     from:
       kind: DockerImage
       name: quay.io/opendatahub/workbench-images:jupyter-datascience-ubi9-python-3.9-2023a-weekly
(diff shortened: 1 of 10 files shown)


VaishnaviHire (approved):
/lgtm

VaishnaviHire:
@atheo89 Expected failure in tests due to outdated tags in test Notebooks.

You would need to update [image values](https://github.com/opendatahub-io/odh-manifests/blob/master/tests/resources/notebook-controller/notebooks/cuda-jupyter-minimal-ubi8-python-3-8.yaml#L13) for all test Notebooks to match the new tags

Issue #752, opened 2023-03-20
Rethink Metrics monitoring Stack in ODH


shalberd:
**Is your feature request related to a problem? Please describe.**
Sicne Openshift 4.6, I believe, there has been a new feature in both OKD and OCP called "Monitoring for User Defined Projects".
Enabling it on a cluster leads to all non-kube* and non-openshift* Namespaces being monitored by a different prometheus in openshift-user-workload-monitoring namespace. At the same time, application metrics timeseries from ServiceMonitors and PodMonitors, as well as kube-state-metrics container- and pod and PVC metrics are available per-namespace, nicely separated by namespace, with its own RBAC. 

The only metrics that cannot be retrieved this way are node-exporter node-level metrics.

The thing is, Red Hat does not recommend mixing own implementations of prometheus operators (we did that on OCP 3.11 in the past and pre-OCP-4.6) with Monitoring for User Defined Projects.

**Describe the solution you'd like**

![Bildschirmfoto 2023-03-20 um 15 10 59](https://user-images.githubusercontent.com/21118431/226390591-5ffed190-c2ba-472a-a884-5ecbf2f1ceaf.png)

With Openshift Monitoring for User Defined Projects, the bringing-in / federation of cluster-level metrics from the kube-state-metrics exporter (pod container restarts , oom, all that stuff) happens automatically at the namespace-level. The only thing not accessible are node-level (node exporter) metrics. Meaning I get such things as kube_pod_container_restarts without an explicit federation servicemonitor. Could you make the federation servicemonitor and prometheus and prometheus operator optional via an overlay in model-mesh? That way, you'd have all the metrics gathering still in there, while making it possible for users who have monitoring for user defined projects enabled to skip the prometheus part and cluster metrics federation part.

The monitoring of the metrics from [odh-model-controller ServiceMonitor](https://github.com/opendatahub-io/odh-manifests/blob/17d5e9084b4e013c12f5340ba269327d74bbf4b5/model-mesh/overlays/odh-model-controller/prometheus/monitor.yaml) works with Monitoring for User Defined Projects, too, by the way.

That is, the [section with the custom monitoring implementation for model mesh could be removed](https://github.com/opendatahub-io/odh-manifests/blob/master/kfdef/odh-core.yaml#L57) from odh-core, as it is achieved with Monitoring for User Defined Projects.

**Describe alternatives you've considered**

**Additional context**

VannTen:
[...]

PR #786, opened 2023-04-12
Modular monitoring stack with overlays

VannTen:
## Description

This tries to implement the idea of using overlays to have a modular monitoring stack in ODH.
The idea comes from #752 , and this is a really early draft to see what it would like.

I'm first trying to define the structure then I'll check the stuff related to model-mesh.


VannTen:
@LaVLaS 
@shalberd 
@anishasthana
I'm interested in what you think

PR #787, opened 2023-04-13
Fix tests to support changes introduced by 1.5 Operator

VaishnaviHire:
## Description
- The deployment name for the operator is updated to `opendatahub-operator-controller-manager`. Update the tests to use the new name.
- DSP Operator tests are failing because, OpenShift Pipelines Operator is being installed in the namespace other than the KfDef namespace. This is not supported, since we have added ownerreferences to KfDef

## Merge criteria:

### DIFF
--- a/tests/Dockerfile
+++ b/tests/Dockerfile
@@ -47,7 +47,7 @@ ENV PATH="${PATH}:$HOME/.local/bin"
 RUN cd $HOME/src/ods-ci && poetry install
     
 
-COPY setup/operatorsetup scripts/install.sh scripts/installandtest.sh $HOME/peak/
+COPY setup/operatorsetup setup/pipelines-op-setup scripts/install.sh scripts/installandtest.sh $HOME/peak/
 COPY resources $HOME/peak/operator-tests/odh-manifests/resources
 COPY util $HOME/peak/operator-tests/odh-manifests
 COPY setup/odh-core.yaml $HOME/kfdef/
--- a/tests/basictests/ds-pipelines.sh
+++ b/tests/basictests/ds-pipelines.sh
@@ -63,13 +63,7 @@ function setup_monitoring() {
 
 function test_metrics() {
     header "Checking metrics for total number of runs, should be 1 since we have spun up 1 run"
-    ## On OCP 4.11, get-token is removed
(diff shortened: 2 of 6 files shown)


harshad16 (approved):
/lgtm
/approve
thanks :+1:

VaishnaviHire:
@LaVLaS Can you review this. This will unblock the failing `odh-manifests` tests

anishasthana (approved):

LaVLaS:
@VaishnaviHire Since the openshift-pipelines installation is just an OLM subscription, we can simplify the workflow by apllying the openshift-pipelines subscription as part of the [tests/scripts/install.sh](https://github.com/opendatahub-io/odh-manifests/blob/master/tests/scripts/install.sh#L10) script and delete the openshift-pipelines test since that was from the older ODH releases where we bundled openshift-pipelines as a component

If you this PR will pass right now then we can keep the current working code and I will submit a follow-up PR to update that workflow

VaishnaviHire:
> @VaishnaviHire Since the openshift-pipelines installation is just an OLM subscription, we can simplify the workflow by apllying the openshift-pipelines subscription as part of the [tests/scripts/install.sh](https://github.com/opendatahub-io/odh-manifests/blob/master/tests/scripts/install.sh#L10) script and delete the openshift-pipelines test since that was from the older ODH releases where we bundled openshift-pipelines as a component
> 
> If you this PR will pass right now then we can keep the current working code and I will submit a follow-up PR to update that workflow

Yes, agreed. I will update the PR to directly use the subscription.

VaishnaviHire:
Failed due to issues creating test clusters.

/test odh-manifests-e2e

/test 411-odh-manifests-e2e

VaishnaviHire:
Expected error in `ci/prow/odh-manifests-e2e` for `4.12` cluster.

`oc get-token..`  is depricated. DS Pipelines [tests](https://github.com/opendatahub-io/odh-manifests/blob/master/tests/basictests/ds-pipelines.sh#L69) need to be updated to fix this. I think we just need to remove the if condition in the linked test. I will can add additional commit in this PR.

anishasthana (approved):
/lgtm

VaishnaviHire:
Merging this since the tests passed

PR #792, opened 2023-04-17
Modular monitoring stack with overlays

VannTen:
## Description

This tries to implement the idea of using overlays to have a modular monitoring stack in ODH.
The idea comes from opendatahub-io/opendatahub-community#115 , and this is a really early draft to see what it would like.

I'm first trying to define the structure then I'll check the stuff related to model-mesh.

(this is the same as opendatahub-io/odh-manifests#786 but changing branch since permissions have moved around)


VannTen:
@anishasthana @LaVLaS @shalberd

anishasthana:
cc @VedantMahabaleshwarkar

Issue #829, opened 2023-05-24
Notebook Images are failing to pull


anishasthana:
We are seeing the following errors on most notebook images when we try to spin up a notebook server using the latest version (the 2023.1 tag) of notebook image. Using an older version of the image is working as expected.

```
failed to pull image "image-registry.openshift-image-registry.svc:5000/odh-applications/jupyter-minimal-notebook:2023.1": rpc error: code = Unknown desc = reading manifest 2023.1 in image-registry.openshift-image-registry.svc:5000/odh-applications/jupyter-minimal-notebook: unknown: unable to pull manifest from [quay.io/opendatahub/workbench-images@sha256:a2a96b9bd8d1aeae20a47197a9ed493a07c902fc0a3981d6109bb3659959c677](http://quay.io/opendatahub/workbench-images@sha256:a2a96b9bd8d1aeae20a47197a9ed493a07c902fc0a3981d6109bb3659959c677): manifest unknown: manifest unknown
```

cc @harshad16

harshad16:
The issue is due to the floating tag `weekly` which changes every week.

If the user installed the cluster a week before the old tag would not be available on the quay.
due to the floating tag and the image digest of the previous being orphaned and no longer available causing the issue of PullBack.

Temporary fix for this would be enable tag being up-to-date with external tag.
Permanent fix would be done via : https://github.com/opendatahub-io/notebooks/issues/103

shalberd:
Yes, they key thing is that the old digest is not present anymore for a weekly build tag such as ` jupyter-datascience-ubi9-python-3.9-2023a-weekly`

@harshad16 
Regarding "old tag would not be available on the quay":

The tag itself is always there, meaning the tag itself at quai io for -weekly does not change
` jupyter-datascience-ubi9-python-3.9-2023a-weekly`
 It is the image digest that changes. 

@VaishnaviHire  I made an error in thinking with respect to long-running notebooks when I assumed that the old digest would still be there in the internal openshift registry for such a tag, did I? ...

harshad16:
@shalberd Yes, I meant to write `old tag digest`. 
as you pointed the tag remains, the digested is not there anymore, that is what I meant with my comment.


About the internal registry, i think it would depend upon the cluster configuration as well right?

shalberd:
@harshad16 regarding "I made an error in thinking with respect to long-running notebooks when I assumed that the old digest / old image version would still be there in the internal openshift registry"

[...]

PR #830, opened 2023-05-24
Allow workbenches to keep the image digest up-to-date

harshad16:
Allow workbenches to keep the image digest up-to-date

## Description

Related-to: https://github.com/opendatahub-io/odh-manifests/issues/829

Openshit allow periodic pull of image digest from it external source 
https://docs.openshift.com/container-platform/4.12/openshift_images/image-streams-manage.html#images-imagestream-import_image-streams-managing
https://docs.openshift.com/container-platform/3.11/dev_guide/managing_images.html#importing-tag-and-image-metadata

with the `scheduled` flag , if the user has out dated sha of the weekly tag
this would help out.

## How Has This Been Tested?

Setup the ODH cluster
Try out the new image tag and update the tag
scheduled will pull the latest tag.

## Merge criteria:


atheo89 (commented):
Nice catch, I think `scheduled` is what we need here
/lgtm

shalberd:
Yup: https://github.com/opendatahub-io/kubeflow/pull/72#issuecomment-1496946371

shalberd:
suggestion:

For non-dev purposes, keep the scheduled: true in the weekly tags out of the imagestreams. That is, keep that out of any branch other than master, e.g. v1.6

On the other hand, on airgapped clusters, scheduled: true will not work anyways with regards to looking at the external registry for changes to the weekly data science image digests ...

notebook-images/base/jupyter-minimal-gpu-notebook-imagestream.yaml
```diff
@@ -23,6 +23,8 @@ spec:
       kind: DockerImage
       name: quay.io/opendatahub/workbench-images:cuda-jupyter-minimal-ubi9-python-3.9-2023a-weekly
```
shalberd, line 24: very good idea to keep the from.name image reference to the source-repo tag-based on the master branch.

notebook-images/base/jupyter-minimal-gpu-notebook-imagestream.yaml
```diff
@@ -23,6 +23,8 @@ spec:
...
     name: "2023.1-cuda-11.8"
+    importPolicy:
+      scheduled: true
     referencePolicy:
       type: Local
```
shalberd, line 29: change referencePolicy to `type: Source `to be independent of the internal openshift registry. I think we can safely assume quay.io will always be available :-)

notebook-images/base/jupyter-minimal-gpu-notebook-imagestream.yaml
```diff
@@ -23,6 +23,8 @@ spec:
       kind: DockerImage
       name: quay.io/opendatahub/workbench-images:cuda-jupyter-minimal-ubi9-python-3.9-2023a-weekly
     name: "2023.1-cuda-11.8"
+    importPolicy:
+      scheduled: true
```
shalberd, line 27: yes, that setting for the weekly-changing tags assures that the imagestream status-section status.tags[x].items.dockerImageReference always refers to the latest digest value, as well as status.tags[x].items.image.

PR #838, opened 2023-06-05
Add KServe manifests to odh-manifests

ReToCode:
## Description
This PR adds the initial manifests for KServe and necessary patches to make it run in ODH. It fetches the manifests from opendatahub-io/kserve and adds some final patches. So updates should be fairly simple in the future.

## How Has This Been Tested?
It was tested using https://github.com/ReToCode/knative-kserve#installation-of-knative-with-service-mesh.

Use then use:
* [kfdef](https://github.com/ReToCode/knative-kserve/blob/main/rhods/kserve-kfdef.yaml)
* [serving runtime](https://github.com/ReToCode/knative-kserve/blob/main/rhods/serving-runtime.yaml)

And test an `InferenceService`
```bash
cat <<-EOF | oc apply -f -
apiVersion: "serving.kserve.io/v1beta1"
kind: "InferenceService"
metadata:
  name: "sklearn-iris"
  namespace: kserve-demo
  annotations:
    sidecar.istio.io/inject: "true"
    sidecar.istio.io/rewriteAppHTTPProbers: "true"
    serving.knative.openshift.io/enablePassthrough: "true"
spec:
  predictor:
    model:
      runtime: kserve-sklearnserver
      modelFormat:
        name: sklearn
      storageUri: "gs://kfserving-examples/models/sklearn/1.0/model"
EOF

oc get isvc -n kserve-demo
NAME           URL                                                                                          READY   PREV   LATEST   PREVROLLEDOUTREVISION   LATESTREADYREVISION            AGE
sklearn-iris   http://sklearn-iris-kserve-demo.apps.rlehmann-ocp-4-12.serverless.devcluster.openshift.com   True           100                              sklearn-iris-predictor-00001   2m14s
```

## Merge criteria:
- [X] The commits are squashed in a cohesive manner and have meaningful messages.
- [X] Testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).
- [X] The developer has manually tested the changes and verified that the changes work

### DIFF
--- a/kserve/README.md
+++ b/kserve/README.md
@@ -1,0 +1,137 @@
+# KServe
+
+KServe comes with one component:
+
+1. [KServe](#KServe)
+
+
+## KServe
+
+Contains deployment manifests for the KServe controller.
+
+- [kserve-controller](https://github.com/opendatahub-io/kserve)
+  - Forked upstream kserve/kserve repository
+
+
+## KServe Architecture
+
+A complete architecture can be found at https://kserve.github.io/website/0.10/modelserving/control_plane.
+
+KServe Control Plane: Responsible for reconciling the InferenceService custom resources. It creates the Knative serverless deployment for predictor, transformer, explainer to enable autoscaling based on incoming request workload including scaling down to zero when no traffic is received. When raw deployment mode is enabled, control plane creates Kubernetes deployment, service, ingress, HPA.
+
+
+## Original manifests
+
+> ❗️Note: Unfortunately, `kfctl` used an outdated version of `kustomize` which cannot process some fields in the KServe manifests.  
+> Thus, we are pre-building the KServe manifests to the [kserve-built](./kserve-built) folder. The [patches](#patches) then use
+> those pre-built KServe manifests to apply our kustomizations.
+
+KServe also uses `kustomize` so we can (indirectly) use [their manifests](https://github.com/opendatahub-io/kserve/tree/master/config).
+
+* `default` is the entrypoint for CRDs, KServe controller and RBAC resources.
+
+The [pre-built KServe manifests](./kserve-built/kserve-built.yaml) are directly referenced in [base/kustomization.yaml](./base/kustomization.yaml).
+
+
+### Updating the manifests
+
+Run the script in [hack](./hack) to manually update the pre-built manifests from the upstream manifests. [this file](./hack/kustomization.yaml) defines the version that is being used:
+
+```bash
+hack/build-kserve-manifests.sh
+```
+```text
+Building KServe manifests
+KServe manifests fetched from upstream and assembled into /odh-manifests/kserve/kserve-built/kserve-built.yaml
+```
+
+
+## Patches
+
+There are patches defined in [kserve/base](./base) to update images and tags for the KServe resources.
+ 
+
+## Installation process
+
+Following are the steps to install Model Mesh as a part of OpenDataHub install:
+
+1. Install the OpenDataHub operator.
+2. Make sure you install Service Mesh and Serverless components and configure them appropriately.
+See [OCP official instructions](https://docs.openshift.com/serverless/1.29/integrations/serverless-ossm-setup.html) and the 
+3. related documentation for [KServe on Openshift](https://github.com/kserve/kserve/blob/master/docs/OPENSHIFT_GUIDE.md#installation-with-service-mesh) from the kserve repo for more.
+4. Create a KfDef that includes the KServe components and runtimes.
+
+```
+apiVersion: kfdef.apps.kubeflow.org/v1
+kind: KfDef
+metadata:
+  name: opendatahub
+  namespace: opendatahub
+spec:
+  applications:
+    - kustomizeConfig:
+        repoRef:
+          name: manifests
+          path: odh-common
+      name: odh-common
+    - kustomizeConfig:
+        repoRef:
+          name: manifests
+          path: kserve
+      name: kserve
+  repos:
+    - name: manifests
+      uri: https://api.github.com/repos/opendatahub-io/odh-manifests/tarball/master
+  version: master
+```
+
+5. You can now create a new project.
+
+6. Make sure that you have a runtime defined in your target namespace (you can use a [template](https://github.com/opendatahub-io/odh-dashboard/blob/main/manifests/modelserving/ovms-ootb.yaml) in ODH).
+
+```yaml
+apiVersion: serving.kserve.io/v1alpha1
+kind: ServingRuntime
+metadata:
+  name: example-runtime
+spec:
+...
+```
+More information in the [KServe docs](https://kserve.github.io/website/0.10/modelserving/servingruntimes/).
+
+7. Create an `InferenceService` CR in your target namespace.
+
+
+## Using KServe in ODH
+
+You can use the `InferenceService` examples from KServe. Make sure to include the additional annotation for OpenShift Service Mesh:
+
+```yaml
+metadata:
+  annotations:
+    sidecar.istio.io/inject: "true"
+    sidecar.istio.io/rewriteAppHTTPProbers: "true"
+    serving.knative.openshift.io/enablePassthrough: "true"
+```
+
+Example:
+
+```yaml
+apiVersion: "serving.kserve.io/v1beta1"
+kind: "InferenceService"
+metadata:
+  name: "sklearn-iris"
+  namespace: kserve-demo
+  annotations:
+    sidecar.istio.io/inject: "true"
+    sidecar.istio.io/rewriteAppHTTPProbers: "true"
+    serving.knative.openshift.io/enablePassthrough: "true"
+spec:
+  predictor:
+    model:
+      runtime: <your-runtime>
+      modelFormat:
+        name: sklearn
+      storageUri: "gs://kfserving-examples/models/sklearn/1.0/model"
+```
+
--- a/kserve/base/kustomization.yaml
+++ b/kserve/base/kustomization.yaml
@@ -1,0 +1,79 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+  - ../kserve-built
+
+namespace: opendatahub
+
+patches:
+  - path: kserve-controller-manager-patch.yaml
+  - path: inferenceservice-config-patch.yaml
+
+configMapGenerator:
+- envs:
+  - params.env
+  name: kserve-parameters
+
+generatorOptions:
+  disableNameSuffixHash: true
+
+vars:
+- fieldref:
+    fieldpath: data.kserve-controller
+  name: kserve-controller
+  objref:
+    apiVersion: v1
+    kind: ConfigMap
+    name: kserve-parameters
+- fieldref:
+    fieldpath: data.kserve-alibi-explainer
+  name: kserve-alibi-explainer
+  objref:
+    apiVersion: v1
+    kind: ConfigMap
+    name: kserve-parameters
+- fieldref:
+    fieldpath: data.kserve-alibi-explainer-version
+  name: kserve-alibi-explainer-version
+  objref:
+    apiVersion: v1
+    kind: ConfigMap
+    name: kserve-parameters
+- fieldref:
+    fieldpath: data.kserve-art-explainer
+  name: kserve-art-explainer
+  objref:
+    apiVersion: v1
+    kind: ConfigMap
+    name: kserve-parameters
+- fieldref:
+    fieldpath: data.kserve-art-explainer-version
+  name: kserve-art-explainer-version
+  objref:
+    apiVersion: v1
+    kind: ConfigMap
+    name: kserve-parameters
+- fieldref:
+    fieldpath: data.kserve-storage-initializer
+  name: kserve-storage-initializer
+  objref:
+    apiVersion: v1
+    kind: ConfigMap
+    name: kserve-parameters
+- fieldref:
+    fieldpath: data.kserve-agent
+  name: kserve-agent
+  objref:
+    apiVersion: v1
+    kind: ConfigMap
+    name: kserve-parameters
+- fieldref:
+    fieldpath: data.kserve-router
+  name: kserve-router
+  objref:
+    apiVersion: v1
+    kind: ConfigMap
+    name: kserve-parameters
+
+configurations:
+  - params.yaml
--- a/kserve/base/params.env
+++ b/kserve/base/params.env
@@ -1,0 +1,8 @@
+kserve-controller=quay.io/opendatahub/kserve-controller:v0.10.2
+kserve-alibi-explainer-version=v0.10.2
+kserve-alibi-explainer=quay.io/opendatahub/kserve-alibi-explainer
+kserve-art-explainer-version=latest
+kserve-art-explainer=quay.io/opendatahub/kserve-art-explainer
+kserve-agent=quay.io/opendatahub/kserve-agent:v0.10.2
+kserve-router=quay.io/opendatahub/kserve-router:v0.10.2
+kserve-storage-initializer=quay.io/opendatahub/kserve-storage-initializer:v0.10.2
--- a/kserve/hack/build-kserve-manifests.sh
+++ b/kserve/hack/build-kserve-manifests.sh
@@ -1,0 +1,19 @@
+#!/usr/bin/env bash
+
+# This script is used to fetch KServe manifests from github.com/opendatahub-io/kserve
+# and bundles them into one big assembled `kserve-built-yaml` file.
+# To update the version, update `hack/kustomization.yaml` and re-run the script with:
+#
+# $ hack/build-kserve-manifests.sh
+
+set -Eeuo pipefail
+
+echo "Building KServe manifests"
+script_dir=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+output_dir="$script_dir/../kserve-built"
+
+command -v kustomize >/dev/null 2>&1 || echo >&2 "kustomize is not installed. Please install kustomize in order to proceed"
+
+kustomize build "$script_dir" > "$output_dir"/kserve-built.yaml
+
+echo "KServe manifests fetched from upstream and assembled into $output_dir/kserve-built.yaml"
--- a/kserve/kserve-built/kserve-built.yaml
+++ b/kserve/kserve-built/kserve-built.yaml
@@ -1,0 +1,18681 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  labels:
+    app.kubernetes.io/part-of: kserve
+    control-plane: kserve-controller-manager
+    controller-tools.k8s.io: "1.0"
+    istio-injection: disabled
+  name: kserve
+---
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    controller-gen.kubebuilder.io/version: v0.4.0
+  creationTimestamp: null
+  labels:
+    app.kubernetes.io/part-of: kserve
+  name: inferencegraphs.serving.kserve.io
+spec:
+  group: serving.kserve.io
+  names:
+    kind: InferenceGraph
+    listKind: InferenceGraphList
+    plural: inferencegraphs
+    shortNames:
+    - ig
+    singular: inferencegraph
+  scope: Namespaced
+  versions:
+  - additionalPrinterColumns:
+    - jsonPath: .status.url
+      name: URL
+      type: string
+    - jsonPath: .status.conditions[?(@.type=='Ready')].status
+      name: Ready
+      type: string
+    - jsonPath: .metadata.creationTimestamp
+      name: Age
+      type: date
+    name: v1alpha1
+    schema:
+      openAPIV3Schema:
+        properties:
+          apiVersion:
+            type: string
+          kind:
+            type: string
+          metadata:
+            type: object
+          spec:
+            properties:
+              affinity:
+                properties:
+                  nodeAffinity:
+                    properties:
+                      preferredDuringSchedulingIgnoredDuringExecution:
+                        items:
+                          properties:
+                            preference:
+                              properties:
+                                matchExpressions:
+                                  items:
+                                    properties:
+                                      key:
+                                        type: string
+                                      operator:
+                                        type: string
+                                      values:
+                                        items:
+                                          type: string
+                                        type: array
+                                    required:
+                                    - key
+                                    - operator
+                                    type: object
+                                  type: array
+                                matchFields:
+                                  items:
+                                    properties:
+                                      key:
+                                        type: string
+                                      operator:
+                                        type: string
+                                      values:
+                                        items:
+                                          type: string
+                                        type: array
+                                    required:
+                                    - key
+                                    - operator
+                                    type: object
+                                  type: array
+                              type: object
+                            weight:
+                              format: int32
+                              type: integer
+                          required:
+                          - preference
+                          - weight
+                          type: object
+                        type: array
+                      requiredDuringSchedulingIgnoredDuringExecution:
+                        properties:
+                          nodeSelectorTerms:
+                            items:
+                              properties:
+                                matchExpressions:
+                                  items:
+                                    properties:
+                                      key:
+                                        type: string
+                                      operator:
+                                        type: string
+                                      values:
+                                        items:
+                                          type: string
+                                        type: array
+                                    required:
+                                    - key
+                                    - operator
+                                    type: object
+                                  type: array
+                                matchFields:
+                                  items:
+                                    properties:
+                                      key:
+                                        type: string
+                                      operator:
+                                        type: string
+                                      values:
+                                        items:
+                                          type: string
+                                        type: array
+                                    required:
+                                    - key
+                                    - operator
+                                    type: object
+                                  type: array
+                              type: object
+                            type: array
+                        required:
+                        - nodeSelectorTerms
+                        type: object
+                    type: object
+                  podAffinity:
+                    properties:
+                      preferredDuringSchedulingIgnoredDuringExecution:
+                        items:
+                          properties:
+                            podAffinityTerm:
+                              properties:
+                                labelSelector:
+                                  properties:
+                                    matchExpressions:
+                                      items:
+                                        properties:
+                                          key:
+                                            type: string
+                                          operator:
+                                            type: string
+                                          values:
+                                            items:
+                                              type: string
+                                            type: array
+                                        required:
+                                        - key
+                                        - operator
+                                        type: object
+                                      type: array
+                                    matchLabels:
+                                      additionalProperties:
+                                        type: string
+                                      type: object
+                                  type: object
+                                namespaceSelector:
+                                  properties:
+                                    matchExpressions:
+                                      items:
+                                        properties:
+                                          key:
+                                            type: string
+                                          operator:
+                                            type: string
+                                          values:
+                                            items:
+                                              type: string
+                                            type: array
+                                        required:
+                                        - key
+                                        - operator
+                                        type: object
+                                      type: array
+                                    matchLabels:
+                                      additionalProperties:
+                                        type: string
+                                      type: object
+                                  type: object
+                                namespaces:
+                                  items:
+                                    type: string
+                                  type: array
+                                topologyKey:
+                                  type: string
+                              required:
+                              - topologyKey
+                              type: object
+                            weight:
+                              format: int32
+                              type: integer
+                          required:
+                          - podAffinityTerm
+                          - weight
+                          type: object
+                        type: array
+                      requiredDuringSchedulingIgnoredDuringExecution:
+                        items:
+                          properties:
+                            labelSelector:
+                              properties:
+                                matchExpressions:
+                                  items:
+                                    properties:
+                                      key:
+                                        type: string
+                                      operator:
+                                        type: string
+                                      values:
+                                        items:
+                                          type: string
+                                        type: array
+                                    required:
+                                    - key
+                                    - operator
+                                    type: object
+                                  type: array
+                                matchLabels:
+                                  additionalProperties:
+                                    type: string
+                                  type: object
+                              type: object
+                            namespaceSelector:
+                              properties:
+                                matchExpressions:
+                                  items:
+                                    properties:
+                                      key:
+                                        type: string
+                                      operator:
+                                        type: string
+                                      values:
+                                        items:
+                                          type: string
+                                        type: array
+                                    required:
+                                    - key
+                                    - operator
+                                    type: object
+                                  type: array
+                                matchLabels:
+                                  additionalProperties:
+                                    type: string
+                                  type: object
+                              type: object
+                            namespaces:
+                              items:
+                                type: string
+                              type: array
+                            topologyKey:
+                              type: string
+                          required:
+                          - topologyKey
+                          type: object
+                        type: array
+                    type: object
+                  podAntiAffinity:
+                    properties:
+                      preferredDuringSchedulingIgnoredDuringExecution:
+                        items:
+                          properties:
+                            podAffinityTerm:
+                              properties:
+                                labelSelector:
+                                  properties:
+                                    matchExpressions:
+                                      items:
+                                        properties:
+                                          key:
+                                            type: string
+                                          operator:
+                                            type: string
+                                          values:
+                                            items:
+                                              type: string
+                                            type: array
+                                        required:
+                                        - key
+                                        - operator
+                                        type: object
+                                      type: array
+                                    matchLabels:
+                                      additionalProperties:
+                                        type: string
+                                      type: object
+                                  type: object
+                                namespaceSelector:
+                                  properties:
+                                    matchExpressions:
+                                      items:
+                                        properties:
+                                          key:
+                                            type: string
+                                          operator:
+                                            type: string
+                                          values:
+                                            items:
+                                              type: string
+                                            type: array
+                                        required:
+                                        - key
+                                        - operator
+                                        type: object
+                                      type: array
+                                    matchLabels:
+                                      additionalProperties:
+                                        type: string
+                                      type: object
+                                  type: object
+                                namespaces:
+                                  items:
+                                    type: string
+                                  type: array
+                                topologyKey:
+                                  type: string
+                              required:
+                              - topologyKey
+                              type: object
+                            weight:
+                              format: int32
+                              type: integer
+                          required:
+                          - podAffinityTerm
+                          - weight
+                          type: object
+                        type: array
+                      requiredDuringSchedulingIgnoredDuringExecution:
+                        items:
+                          properties:
+                            labelSelector:
+                              properties:
+                                matchExpressions:
+                                  items:
+                                    properties:
+                                      key:
+                                        type: string
+                                      operator:
+                                        type: string
+                                      values:
+                                        items:
+                                          type: string
+                                        type: array
+                                    required:
+                                    - key
+                                    - operator
+                                    type: object
+                                  type: array
+                                matchLabels:
+                                  additionalProperties:
+                                    type: string
+                                  type: object
+                              type: object
+                            namespaceSelector:
+                              properties:
+                                matchExpressions:
+                                  items:
+                                    properties:
+                                      key:
+                                        type: string
+                                      operator:
+                                        type: string
+                                      values:
+                                        items:
+                                          type: string
+                                        type: array
+                                    required:
+                                    - key
+                                    - operator
+                                    type: object
+                                  type: array
+                                matchLabels:
+                                  additionalProperties:
+                                    type: string
+                                  type: object
+                              type: object
+                            namespaces:
+                              items:
+                                type: string
+                              type: array
+                            topologyKey:
+                              type: string
+                          required:
+                          - topologyKey
+                          type: object
+                        type: array
+                    type: object
+                type: object
+              nodes:
+                additionalProperties:
+                  properties:
+                    routerType:
+                      enum:
+                      - Sequence
+                      - Splitter
+                      - Ensemble
+                      - Switch
+                      type: string
+                    steps:
+                      items:
+                        properties:
+                          condition:
+                            type: string
+                          data:
+                            type: string
+                          name:
+                            type: string
+                          nodeName:
+                            type: string
+                          serviceName:
+                            type: string
+                          serviceUrl:
+                            type: string
+                          weight:
+                            format: int64
+                            type: integer
+                        type: object
+                      type: array
+                  required:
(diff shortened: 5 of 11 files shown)


skonto:
@ReToCode nice! I think the next step would be to have some basic testing (smoke tests) in https://github.com/opendatahub-io/odh-manifests/tree/master/tests/basictests? I am not aware of the CI work for ODH and KServe but it would be nice to be able to test the manifests and future work against Kserve examples etc.

Jooho:
@ReToCode Opendatahub operator is using kfctl and it is using the base folder as an entry point. I can help to change the structure a little bit to be worked by kfctl .
However, there is another issue. kfctl is not updated for a long time while kustomize keeps updating. So some parameters such as `replacements` in the kserve/default/kustomization.yaml does not seem that it is possible to be parsed by kfcfl. So even though we change the directory structure, it should not work at the moment.

 @VaishnaviHire do we have the plan to support kustomize for the odh-manifests?

kserve/odh-overlays/controller/inferenceservice-config-patch.yaml
```diff
@@ -0,0 +1,60 @@
...
+        "cpuLimit": "1"
+    }
+  router: |-
+    {
+        "image" : "$(kserve-router)",
+        "memoryRequest": "100Mi",
+        "memoryLimit": "1Gi",
+        "cpuRequest": "100m",
+        "cpuLimit": "1"
+    }
```
Jooho, line 60: This will always overwrite the original file because it is 1 json object. So I think it would be better to update original file for `the image part`
For example, if router param in the original file changed like following 
~~~
  router: |-
    {
       "test":"test",  <---
        "image" : "$(kserve-router)",
        "memoryRequest": "100Mi",
        "memoryLimit": "1Gi",
        "cpuRequest": "100m",
        "cpuLimit": "1"
    }
~~~
It will not be applied because this file overwrite the one.

kserve/README.md
```diff
@@ -0,0 +1,121 @@
+# KServe
+
+KServe comes with two component:
+
+1. [KServe](#KServe)
+2. [KServe runtimes](#KServe-Runtimes)
```
Jooho, line 6: is there a reason to separate this part(kserve runtime) from `kserve`?

kserve/README.md
```diff
@@ -0,0 +1,121 @@
...
+* [runtimes](./odh-overlays/runtimes)
+
+
+### Installation process
+
+Following are the steps to install Model Mesh as a part of OpenDataHub install:
+
+1. Install the OpenDataHub operator
+2. Make sure you install Service Mesh and Serverless components and configure them appropriately
+3. Create a KfDef that includes the KServe components and runtimes
```
Jooho, line 51: have you tested this kfdef with your change?

kserve/README.md
```diff
@@ -0,0 +1,121 @@
...
+    - name: manifests
+      uri: https://api.github.com/repos/opendatahub-io/odh-manifests/tarball/master
+  version: master
+```
+
+4. You can now create a new project and create an `InferenceService` CR.
+
+## Using KServe in ODH
+
+You can use the `InferenceService` examples from KServe. Make sure to include the additional annotation for OpenShift Service Mesh:
```
Jooho, line 81: This kserve manifest assumes that the cluster has istio or service-mesh installed as pre-requisite?

kserve/README.md
```diff
@@ -0,0 +1,121 @@
...
+
+Currently, the target namespace service account must be allowed to run as `anyuid`, so allow this using:
+
+```bash
+oc adm policy add-scc-to-user anyuid -z default -n <your-namespace>
+```
+
+**Reason**
+* for istio: allow to run as user 1337 because of https://istio.io/latest/docs/setup/additional-setup/cni/#compatibility-with-application-init-containers
+* for the python images of KServe: allow to run as user 1000 because of: https://github.com/kserve/kserve/blob/master/python/aiffairness.Dockerfile#L46
```
Jooho, line 121: When a pod is started, this user number should be changed randomly, no?

kserve/odh-overlays/runtimes/kustomization.yaml
Jooho, line 1: Why is this file needed?

kserve/hack/update-kserve-manifests.sh
```diff
@@ -0,0 +1,20 @@
...
+cd "$tmp_dir" || exit
+git remote add -f origin https://github.com/kserve/kserve.git
+git config core.sparseCheckout true
+echo "config" >> .git/info/sparse-checkout
+git pull origin master
+
+rm -rf "$kserve_dir"
+mkdir "$kserve_dir"
+
+mv ./config/* "$kserve_dir"
```
israel-hdez, line 18: Can we refer to the github url rather than doing a hard-copy?
For example, in `kserve/odh-overlays/controller/kustomization.yaml`, rather than having:
```yaml
resources:
  - ../../kserve/default/
```

you could have:
```yaml
resources:
  - github.com/opendatahub-io/kserve/config/default?ref=master
```

(note, I didn't test this, but this docs say it should be possible: https://kubectl.docs.kubernetes.io/references/kustomize/kustomization/resource/)

I don't like too much copying/synching files.

kserve/odh-overlays/controller/inferenceservice-config-patch.yaml
ReToCode: The idea was to keep the `kserve` resources as it to be able to update them via the [script](https://github.com/opendatahub-io/odh-manifests/pull/838/files#diff-80e7482e52299d84ae57e4680cf757973a1049eb83e088fa2c96d49f01b27e43). If we change the original file, it gets a bit harder to update. If we want to keep that approach, we need to replace the whole block here, as I don't think kustomize supports replacing sub-parts of a inline yaml. 

I'm fine with both ways. WDYT?

kserve/README.md
ReToCode: KServe does that, I just followed that. I think it is because the runtimes are cluster-scoped and kustomize, unnecessarility, adds the namespace to all the resources. So. `kserve/base/runtimes/kustomization.yaml` does not define the namespace field.

kserve/README.md
ReToCode: No, I just tested it with the kustomize part, as I do not know how to use it/what the full process is. Also there was the open question about the multiple entrypoints.

kserve/README.md
ReToCode: Yes indirectly. KServe expects Knative to be installed, which depends on a networking layer which could either be Kourier or Istio. As we want to go with Istio (and KServe has that as default), this results in the expectation that both are installed. I think there is still some discussion on how the full setup process would be, right?

kserve/README.md
ReToCode: OpenShift would change the UID yes, but KServe images create folders/files for the user `1000`. In the runtime we then get errors, as the runtime user does not have permissions for files of user `1000`. The user id should be changed in the ODH fork when we build these images ourselves. If this is done, we can drop that part.

kserve/odh-overlays/runtimes/kustomization.yaml
ReToCode: Its here for future overrides. I assumed you would want to overwrite which runtimes we bring to ODH and re-write the images+tags of those. As I did not know which runtimes you are targeting, I omitted these changes for now.

ReToCode:
> Opendatahub operator is using kfctl and it is using the base folder as an entry point. I can help to change the structure a little bit to be worked by kfctl .

I updated the folder structure according to [this](https://redhat-internal.slack.com/archives/C051XSYFT7U/p1685973858778509?thread_ts=1685948337.379939&cid=C051XSYFT7U), then we could have the following:

```yaml
  - kustomizeConfig:
      repoRef:
        name: odh-manifests
        path: kserve/controller
    name: kserve-controller
  - kustomizeConfig:
      repoRef:
        name: odh-manifests
        path: kserve/runtimes
    name: kserve-runtimes
```

Would that work?

kserve/hack/update-kserve-manifests.sh
ReToCode: Awesome! I did not know that, now we can drop 99% of the stuff in this PR :)

danielezonca:
FYI ticket to discuss the inclusion of KServe into ODH
- https://github.com/opendatahub-io/opendatahub-community/issues/54

kserve/README.md
Jooho: OpenDataHub does not support cluster-scoped runtime. Do you know how to create namespace-scoped runtime(ServingRuntime)?  Plus, OpenDataHub modelserving does not create any runtime because the ServingRuntime will be created by ODH console using [this template](https://github.com/opendatahub-io/odh-dashboard/blob/main/manifests/modelserving/ovms-ootb.yaml).  At the moment, we only support ovms.

kserve/README.md
Jooho: Right, I just wanted to know what applications are required before installing these manifests, and I think it's good to have that listed somewhere.

kserve/hack/update-kserve-manifests.sh
Jooho: I don't think this is a good idea.. because upstream kserve/kserve manifests can be changed anytime and definitely it will impact to OpenDataHub. We need to have a stable manifests always.

Jooho:
@ReToCode Technically it should be ok but I don't know why we need 2 entry points.

kserve/hack/update-kserve-manifests.sh
israel-hdez: I think `ref=master` could be replaced by `ref={commitHash}` to make it stable.

I also saw that it is changed to use `opendatahub-io/kserve` rather than upstream. So, upstream becomes less of a concern.

Jooho: That looks better.

kserve/odh-overlays/controller/inferenceservice-config-patch.yaml
israel-hdez: Perhaps a custom transformer: https://kubectl.docs.kubernetes.io/guides/extending_kustomize/. But I'm not sure if we want to use custom transformers.

If not, I don't see other way of doing it. So, I agree this is quite fine for now.

kserve/README.md
ReToCode: Ah, KServe also defines [ServingRuntime](https://kserve.github.io/website/0.10/modelserving/servingruntimes/) instead of the cluster scoped one. So we could omit it here completely and deploy the `ServingRuntime` from the template then.

kserve/README.md
ReToCode: Hm what about going back to having a script that gets the resources from KServe repo, then running `kustomize build` using our overlays and use the generated output as entrypoint for kfctl? Not super clean, but would solve all issues around kustomize without having to modify the KServe resources.
Other ideas? @Jooho, @israel-hdez

kserve/hack/update-kserve-manifests.sh
ReToCode: I'll do the updates, depending on the outcome of the discussion in https://github.com/opendatahub-io/odh-manifests/pull/838#discussion_r1227565608

ReToCode:
> @ReToCode Technically it should be ok but I don't know why we need 2 entry points.

Would have been for the runtimes (the namespace stuff). But if we omit it here, we can drop them.

kserve/controller/odh-overlays/kserve-controller-manager-patch.yaml
```diff
@@ -0,0 +1,21 @@
+apiVersion: apps/v1
```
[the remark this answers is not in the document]
ReToCode: I think it should work. I did follow the pattern from other folders in this repo. I was not sure if you need the variables in the configmap on the cluster for something else. I can also use this approach, depending on the outcome of https://github.com/opendatahub-io/odh-manifests/pull/838#discussion_r1227565608.

kserve/controller/base/kustomization.yaml
```diff
@@ -0,0 +1,70 @@
...
+
+configMapGenerator:
+- envs:
+  - params.env
+  name: kserve-parameters
+
+generatorOptions:
+  disableNameSuffixHash: true
+
+vars:
```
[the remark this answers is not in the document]
ReToCode: Yeah that's a bit of a bummer with `kfctl`.

kserve/hack/update-kserve-manifests.sh
ReToCode: Yeah, that would be resolved if we do https://github.com/opendatahub-io/odh-manifests/pull/838#discussion_r1227565608. Then we'd have the final YAMLs in this repo here. Would that work?

kserve/README.md
israel-hdez: > running `kustomize build` using our overlays and use the generated output as entrypoint for kfctl?

This would be quite fine IMO. Although, I still think we don't need to copy the files. I mean, I think running `kustomize build` will still generate what we need without copying files. Then, I guess the only thing left is to add operator parameters, but this should be a quite small task.

kserve/hack/update-kserve-manifests.sh
israel-hdez: > Then we'd have the final YAMLs in this repo here. Would that work?

That works and I think it also would work with the outdated kfdef, because we would need minimal parametrs for the operator.

kserve/README.md
Jooho: This script will be used for opendatahub and rhods. So we need to images and some parts dynamically. If we generate final manifest by `kustomize build`, we can not reuse this manifests.

ReToCode: @israel-hdez Exactly. I'll update the PR to showcase what I imagined.

ReToCode:
* Updated the PR according to https://github.com/opendatahub-io/odh-manifests/pull/838#discussion_r1228987326
* Added new files to fix an issue with webhook-certificates
* Resolved discussions that are now no longer relevant
* Updated the description of the PR with an updated installation/testing manual
* The two entrypoints are no longer necessary, as runtimes come from the [template](https://github.com/opendatahub-io/odh-manifests/pull/838#discussion_r1226971745). For Kfdef is would be something like now:
```yaml
  - kustomizeConfig:
      repoRef:
        name: odh-manifests
        path: kserve
    name: kserve
```
Still assuming that Serverless + Service Mesh are installed prior to installing ODH.

kserve/odh-overlays/kustomization.yaml
```diff
@@ -0,0 +1,49 @@
...
+      - op: remove
+        path: "/webhooks/0/clientConfig/caBundle"
+    target:
+      kind: ValidatingWebhookConfiguration
+
+  # Inject CA bundle into webhooks
+  - patch: |-
+      - op: add
+        path: "/metadata/annotations/cert-manager.io~1inject-ca-from"
+        value: "opendatahub/serving-cert"
```
ReToCode, line 38: @israel-hdez any idea if we can use the namespace in the base/kustomization.yaml here somehow?

israel-hdez: Hmm, I don't know.

A quick google search lead me to this: https://github.com/kubernetes-sigs/kustomize/issues/603#issue-388571646. The `objref` targets a `Namespace`, but if there isn't one in our case, I think we may use any existent object.

It's just a pointer. I don't know if it is good for our case.

kserve/README.md
israel-hdez: @ReToCode  It is looking good to me. Let's wait for @Jooho thoughts.

Jooho: @LaVLaS @VaishnaviHire what do you think about this approach for RHODS/ODH?

danielezonca:
Issue for the PR (just for tracking) https://github.com/opendatahub-io/odh-manifests/issues/848

kserve/kserve-built/kserve-built.yaml
danielezonca, line 1: As far as I see all the control plane components are going to be deployed in a `kserve` namespace. Is this something that it is fixed or can be overridden? 
I think modelmesh controller is deployed in a specific namespace and I was thinking if it makes sense to use the same for both

kserve/hack/build-kserve-manifests.sh
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env bash
+
+# This script is used to build upstream KServe manifests
+# as kfcli uses an outdated kustomize version which cannot
+# process some features like `replacements` in the KServe manifests.
+#
+# Usage:
+# $ hack/build-kserve-manifests.sh
+
```
rhuss, line 9:
```suggestion
set -euo pipefail

```

recommended for avoid otherwise undetected errors.

rhuss: well, just saw that the script is not that huge :-), so might not be worth (still, I add this to every script by habit)

skonto: I will add `-E` for tracing too, we used it in our scripts.

kserve/README.md
skonto: In order to finalize the discussion my understanding is that we dont use any runtimes as they are clusterscoped and not supported.

kserve/README.md
skonto: @LaVLaS @VaishnaviHire gentle ping.

skonto:
From the base dir  we include the specialization
```
base/kustomization.yam:

resources:
  - ../odh-overlays
```

  
Should not this be  the other way around and run `kustomize build odh-overlays instead` of `kustomize build base`.
That is the common pattern I see: https://github.com/kubernetes-sigs/kustomize/blob/master/examples/helloWorld/README.md#deploy.

kserve/odh-overlays/kustomization.yaml
israel-hdez: Maybe @bartoszmajsak can recommend something here. I think he dealt with something similar.

kserve/README.md
```diff
@@ -0,0 +1,140 @@
...
+
+```yaml
+apiVersion: serving.kserve.io/v1alpha1
+kind: ServingRuntime
+metadata:
+  name: example-runtime
+spec:
+...
+```
+More information in the [KServe docs](https://kserve.github.io/website/0.10/modelserving/servingruntimes/).
```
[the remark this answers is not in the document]
israel-hdez: I think I agree for ODH as a platform.
However, at least for ModelMesh, readmes of most repositories enumerate instructions assuming using the isolated (or almost isolated) component. So, I see this more as dev-focused instructions rather than end-user instructions.

israel-hdez:
> Should not this be the other way around and run `kustomize build kserve/odh-overlays` instead of `kustomize build kserve/base`.

I'm neutral to this, because this is implementation detail.

For testing purposes we can use `kustomize`. But end-users won't use `kustomize`, but will create a `KfDef` which is going to hide the implementation detail.

skonto:
I tried to test with a kdef: 

```apiVersion: kfdef.apps.kubeflow.org/v1
kind: KfDef
metadata:
  name: opendatahub
  namespace: opendatahub
spec:
  applications:
    - kustomizeConfig:
        repoRef:
          name: manifests
          path: odh-common
      name: odh-common
    - kustomizeConfig:
        repoRef:
          name: manifests
          path: kserve
      name: kserve
  repos:
    - name: manifests
      uri: https://api.github.com/repos/ReToCode/odh-manifests/tarball/add-kserve-manifests
```

I got this error:

`(kubeflow.error): Code 500 with message: kfApp Apply failed for kustomize: (kubeflow.error): Code 500 with message: error evaluating kustomization manifest for kserve: accumulating resources: couldn't make target for path 'base': json: unknown field "labels"`

israel-hdez:
> I got this error:
> 
> `(kubeflow.error): Code 500 with message: kfApp Apply failed for kustomize: (kubeflow.error): Code 500 with message: error evaluating kustomization manifest for kserve: accumulating resources: couldn't make target for path 'base': json: unknown field "labels"`

Looks like the operator does not support this: https://github.com/opendatahub-io/odh-manifests/pull/838/files#diff-4cbfedfaefc3006a09fe41992e3acb46b140d9b847b4f86005c1b6307dbd50acR67-R70

Try replace that with this:
```yaml
commonLabels:
  app.kubernetes.io/part-of: kserve
```

skonto:
Thanks @israel-hdez I will try. Btw I just verified again the manifests with the instructions above provided by Reto. The sample Iris service is up and running.

israel-hdez:
> The next pending thing is to solve the problem with rendering all urls and contents based on the namespace of the KDef. I suspect that if I create Kdef with a ns different than `opendatahub` it will fail as some stuff are not parameterized eg. patches svc urls.

@skonto I'm attaohing a patch that you can apply with `git apply ns-patch.txt`. It seems to cover this pending point :-)

[ns-patch.txt](https://github.com/opendatahub-io/odh-manifests/files/11852443/ns-patch.txt)

skonto:
@israel-hdez the patch does not seem to work:

```
apiVersion: admissionregistration.k8s.io/v1
kind: MutatingWebhookConfiguration
metadata:
  annotations:
    cert-manager.io/inject-ca-from: $(NAMESPACE)/serving-cert
  creationTimestamp: "2023-06-26T10:15:35Z"
  generation: 3
  labels:
    app.kubernetes.io/part-of: kserve
```

kserve/README.md
skonto: Issue created: https://github.com/opendatahub-io/kserve/issues/13.

kserve/odh-overlays/kserve-controller-manager-patch.yaml
```diff
@@ -0,0 +1,21 @@
...
+        - image: $(kserve-controller)
+          name: manager
+          env:
+            - name: POD_NAMESPACE
+              valueFrom:
+                fieldRef:
+                  fieldPath: metadata.namespace
+            - name: SECRET_NAME
+              value: kserve-webhook-server-cert
+            - name: ISTIO_WITH_CNI_AND_DNS_PROXY
```
skonto, line 20: According to the latest status of the PR here: https://github.com/kserve/kserve/pull/2853, we support passing the uid via an annotation.:
```
apiVersion: serving.kserve.io/v1beta1
kind: InferenceService
metadata:
  name: sklearn-from-uri
  annotations:
    sidecar.istio.io/inject: "true"
    sidecar.istio.io/rewriteAppHTTPProbers: "true"
    serving.knative.openshift.io/enablePassthrough: "true"
    serving.kserve.io/storage-initializer-uid: "1000700001"
spec:
  predictor:
    model:
```
The operator needs to pass that annotation or the user.

skonto:
@israel-hdez I verified again that the patch does work locally but does not work

skonto:
I tried to do a quick [upgrade](https://github.com/opendatahub-io/opendatahub-operator/compare/master...skonto:opendatahub-operator:test_upgrade_kust?expand=1) but it seems opendatahub operator requires api changes too eg. coordinator_test unit test is broken.

israel-hdez:
> @israel-hdez I verified again that the patch does work locally but does not work when applied by the operator, probably the kustomize lib used there is not up to date?
> 

I think this is a well known fact. I thought the patch was going to work, given it doesn't use any unsupported `kfctl` construct.

> Update: The kustomize v3.3.1 used here: https://github.com/opendatahub-io/opendatahub-operator/blob/master/go.mod#L44 is 4years old and https://github.com/kubernetes-sigs/kustomize/blob/v3.3.1/pkg/transformers/config/defaultconfig/varreference.go does not contain a ref for namespace.

`varreference` can be changed and is being changed in the `kserve/odh-overlays/params.yaml` file. IIUC, this configures _targets_ for replacements; not value/var _sources_ . So, probably it is not relevant that there isn't a ref for namespaces.

---

I think we will need to reach out @VaishnaviHire for advice.

skonto:
> I think we will need to reach out @VaishnaviHire for advice.

Why not upgrade the operator? 

> I think this is a well known fact

It was not to me.

israel-hdez:
> > I think we will need to reach out @VaishnaviHire for advice.
> 
> Why not upgrade the operator at some point?

I think I also would be happy that this happens :-)

That said, AFAIK, the operator will undergo a major rework. So, I think upgrading the operator codebase may not have enough value at the moment.

kserve/kserve-built/kserve-built.yaml
Jooho, line 1: Looks like it has cert-manger cr `Certificate`.  RHODS/ODH do not install cert-manager. Is there a way to use [serving-cert](https://docs.openshift.com/container-platform/4.13/security/certificates/service-serving-certificate.html)

kserve/odh-overlays/kustomization.yaml
danielezonca: @VaishnaviHire @LaVLaS @anishasthana do you have any idea/experience with a similar use case?
We need to inject the namespace value as parameter in these places
https://github.com/skonto/odh-manifests/commit/68210020873d31af27859f336d8ac8725e32e52c

It is possible with a recent version of kustomize but this is not really an option at this stage so I guess if you solved this problem in the past with some different workaround.
We can probably hardcoded a similar "pattern" in the operator codebase itself (like `Replace(rawResource, "$(NAMESPACE)", r.namespace)`) but it is quite a hack :)

bartoszmajsak: From the top of my head, if you don't want to change / hack anything in the source code of the operator, you could define a k8s `batch/v1 Job` as part of your kustomize manifests in which you can apply this patch using regular `kubectl patch`. At this point, you will know your target namespace e.g. `NAMESPACE=$(cat /var/run/secrets/kubernetes.io/serviceaccount/namespace)` and use envsubst in the Job script to replace $NAMESPACE in the patch before applying.

kserve/kserve-built/kserve-built.yaml
skonto: @Jooho It should be possible. I see a couple of options here beyond the cert-manager based one. Some background info first.
Controller runtime expects a specific default value for the mounted dir containing the secret. The default dir is https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/webhook/server.go#L148 /tmp/k8s-webhook-server/serving-certs. KServe does not do anything special other than creating a cert manager certificate that at the end of the day creates the secret that is mounted at that path.  
Here is the pod mounting: https://github.com/kserve/kserve/blob/master/config/manager/manager.yaml#L52. The secret name is `kserve-webhook-server-cert`. The secret data keys must be: tls.crt, tls.key.
The options I see:

1)  - Remove the cert-manager certificate
     - Annotate the webhook svc namely kserve-webhook-server-service with:
      ```  oc annotate service test1 service.beta.openshift.io/serving-cert-secret-name=kserve-webhook-server-cert```
     - Then the secret should be mounted as usual by the KServe manager pod
     - Annotate the mutating, validation webhooks so the API server knows how to validate the webhook server.
     `oc annotate mutatingwebhookconfigurations inferenceservice.serving.kserve.io \
     service.beta.openshift.io/inject-cabundle=true`
     
2) Allow olm to generate the certificates assuming that kserve will by managed by an olm based operator (https://olm.operatorframework.io/docs/advanced-tasks/adding-admission-and-conversion-webhooks/#certificate-authority-requirements).
This is something we use currently for the Serverless operator. This requires to pass some parameters to the ctrl manager as could check here: https://github.com/openshift-knative/serverless-operator/blob/main/knative-operator/cmd/manager/main.go#L106-L110.
OLM mounts automatically the certificates at: 

```
The TLS Cert file will also be mounted to the deployment at /apiserver.local.config/certificates/apiserver.crt
The TLS Key file will also be mounted to the deployment at /apiserver.local.config/certificates/apiserver.key
```
The good thing is that OLM will patch the crds declared in the corresponding csv automatically so no need to annotate anything. This works because you need to declare all your webhooks in the operator's csv: https://github.com/openshift-knative/serverless-operator/blob/main/olm-catalog/serverless-operator/manifests/serverless-operator.clusterserviceversion.yaml#L1142.

This kind of management would require to modify the hookserver setup here: https://github.com/kserve/kserve/blob/master/cmd/manager/main.go#L218.

3) Manually setup the certificate (plus rotation) from the opnedatahub side.

In any case the secret is the api for the controller runtime.

We could try option 1 easily. In general I am in favor to use cert-manager if available and kserve comes with that by default.

skonto: @Jooho 
 here is the working version without the cert-manager:
https://github.com/ReToCode/odh-manifests/compare/test...ReToCode:odh-manifests:test-no-cert?expand=1
Full guide:
https://gist.github.com/skonto/5e3ac71cbb527eed629c8381edf1e425

danielezonca: Very nice @skonto !
This change has also a positive side effect: it removes the part of the manifest where the problematic `namespace` patch was so if we take this path we should be able to proceed with the PR "as is"

israel-hdez:
@ReToCode 

> The images from https://github.com/opendatahub-io/kserve are not yet there

I've created these community images:
* https://quay.io/repository/opendatahub/kserve-controller?tab=tags
* https://quay.io/repository/opendatahub/kserve-router?tab=tags
* https://quay.io/repository/opendatahub/kserve-agent?tab=tags

I was unsure about the explainer and storage-initializer, so I didn't created those. If you need them, I can do it.

Note that I simply built and pushed. I'm unsure if the images actually work.

israel-hdez:
@ReToCode BTW, I built the most recent v0.10.2, while here you are using v0.10.1. If you need the older version, let me know.

ReToCode:
@israel-hdez Thank you!

> I was unsure about the explainer and storage-initializer, so I didn't created those. If you need them, I can do it.

Yes we need them. The explainer is necessary for the [explainability feature](https://kserve.github.io/website/0.10/modelserving/explainer/explainer/#inferenceservice-explainer) in KServe and the storage initializer for getting the model from S3 and other storages. Can you also build those?

> Note that I simply built and pushed. I'm unsure if the images actually work.

As soon as the other two are here, we can try to run them.

> BTW, I built the most recent v0.10.2, while here you are using v0.10.1. If you need the older version, let me know.

That should be fine. Not super sure about the versioning concept here, but I expect the images should be aligned with the manifests from the same version (for now we get them from master anyways). I'll update them as well when the two other images are there.

israel-hdez:
@ReToCode 

Here they are:
* https://quay.io/repository/opendatahub/kserve-alibi-explainer?tab=tags
* https://quay.io/repository/opendatahub/kserve-art-explainer?tab=tags
* https://quay.io/repository/opendatahub/kserve-storage-initializer?tab=tags

Note that `kserve-art-explainer` does not have a `v0.10.2` tag because it wasn't there on that version.

> images should be aligned with the manifests from the same version (for now we get them from master anyways)

So, FYI, `latest` tags are images built from `master` branch. The versioned tags are built from the git tag that I mirrored when building the images: https://github.com/opendatahub-io/kserve/releases/tag/v0.10.2.

kserve/kserve-built/kserve-built.yaml
israel-hdez: @skonto @ReToCode I talked with JooHo and we are good with the version without cert-manager. If you can apply it, I think work left would be doing a sanity check.

israel-hdez:
> (for now we get them from master anyways)

Thinking a little more... shouldn't we, indeed, be basing this PR out of the v0.10.x tag?

ReToCode:
@israel-hdez image references updated and generated the yamls from tag v0.10.2.

@Jooho, @danielezonca what is your idea for OCP specific patches in the KServe fork repo? How do you usually do that?
In our midstream we have git-patches on the fork to apply our changes for multiple tags/branches. As we would need a patch that handles the OpenShift UID which we probably cannot bring upstream. If this patch is in the built image, we could omit the `anyuid` stuff.

kserve/base/params.env
```diff
@@ -0,0 +1,7 @@
+kserve-controller=docker.io/skonto/kserve-manager:dev2
```
skonto, line 1: We can revert when things are properly built with all OCP patches.

ReToCode:
Updated this PR, it is not ready yet, as we need the patches via https://github.com/opendatahub-io/kserve/pull/14 first. 

But, basically:
- The script now just fetches the KServe manifests from opendatahub-io/kserve and expects them to be able to run with kfdef
- PR assumes, that we still want to define the images + tags in odh-manifests, so there are still some odh-overlays left (up for discussion.) WDYT @danielezonca , @Jooho , @israel-hdez ?

kserve/odh-overlays/kserve-controller-manager-patch.yaml
ReToCode: I'll do a patch on the opendatahub-io/kserve repository as soon as https://github.com/kserve/kserve/pull/2853 is in.

kserve/base/params.env
ReToCode: As it is "broken" until we have the correct manifests in opendatahub-io/kserve, I omitted your changes. I hope in the mean time we can get the upstream PR in, so this should no longer be necessary.

kserve/kserve-built/kserve-built.yaml
ReToCode: This is now part of: https://github.com/opendatahub-io/kserve/pull/14.

danielezonca:
> Updated this PR, it is not ready yet, as we need the patches via [opendatahub-io/kserve#14](https://github.com/opendatahub-io/kserve/pull/14) first.
> 
> But, basically:
> 
> * The script now just fetches the KServe manifests from opendatahub-io/kserve and expects them to be able to run with kfdef
> * PR assumes, that we still want to define the images + tags in odh-manifests, so there are still some odh-overlays left (up for discussion.) WDYT @danielezonca , @Jooho , @israel-hdez ?

Yes I think here the manifest is mainly going to do this overlay to specify the versions

ReToCode:
We discussed this yesterday. As the "big" changes are going to be in the kserve fork repo, we can use the individual files here, which makes comparing between versions a lot easier. 

What changes do you have in mind?

israel-hdez:
> We discussed this yesterday. As the "big" changes are going to be in the kserve fork repo, we can use the individual files here, which makes comparing between versions a lot easier.
> 

Interesting. It is probably just me, but yesterday when I was working on the opendatahub-io/kserve PR, I found quite easy to compare two big files, in contrast to writing a script to individually compare the smaller files. Well... I'll let it be.

> What changes do you have in mind?

If this "side" is going to expect that opendatahub-io/kserve manifests work with KfCtl, that's currently not true at the moment. I simply applied OpenShift patches. So, I still need to adapt those manifests -- I guess this would be applying JooHo changes from his branch.

Perhaps, this is the part that makes noise to me. Adapting opendatahub-io/kserve manifests makes them diverge from upstream, and pulling new changes is going to constantly require conflict resolution when upstream manifests change. So, I'm not sure if the tradeoff with easier comparison between versions really pays out. I'm not even sure why we would want to be comparing both sides :-/ But, anyway, if this is what was agreed yesterday, I'll let it be.

israel-hdez:
> intermediate kustomize build to "squash" into a single file

So, I've updated my opendatahub-io/kserve#14 PR still under this assumption.
Please, let me know if that's not going to be the case.

ReToCode:
> seems a good tradeoff for now (aka waiting the new operator)

I'm totally fine with both approaches. This was just what I got from @Jooho short demo. So if that is fine for Jooho as well, I can update it that way 👍

israel-hdez:
@ReToCode @skonto FYI I tried the previous big-file approach with some simplifications (i.e. no overlay): https://github.com/opendatahub-io/odh-manifests/compare/master...israel-hdez:838-kserve-manifests?expand=1

Apparently, the ODH operator is happy with it.

ReToCode:
> FYI I tried the previous big-file approach with some simplifications (i.e. no overlay):

Yeah, in the end it is just a question of preference. I originally followed the pattern in `model-mesh`. But your approach is quite simple. I updated my PR accordingly.

We should be able to test everything end-to-end when your PR on KServe and the `anyuid` patch is merged & built.

kserve/base/kustomization.yaml
```diff
@@ -0,0 +1,79 @@
...
+- fieldref:
+    fieldpath: data.kserve-router
+  name: kserve-router
+  objref:
+    apiVersion: v1
+    kind: ConfigMap
+    name: kserve-parameters
+
+commonLabels:
+  app.kubernetes.io/part-of: kserve
```
israel-hdez, line 79: I think this should be:

```suggestion
configurations:
  - params.yaml
```

Otherwise, the params are not being applied. 
Also, this common label [is already applied on the other side](https://github.com/opendatahub-io/kserve/blob/release-0.10/config/default/kustomization.yaml#L7-L9).

israel-hdez (changes requested):
@ReToCode Can you also, please, squash the commits?

ReToCode:
> @ReToCode Can you also, please, squash the commits?

Done and tested, see https://redhat-internal.slack.com/archives/C05742W6F7T/p1689149577193889.

ReToCode:
> oc apply -f ./odh-core.yaml
> error: resource mapping not found for name: "odh-core" namespace: "opendatahub" from "./odh-core.yaml": no matches for kind "KfDef" in version "kfdef.apps.kubeflow.org/v1"

Pretty sure this is related to my changes.

/test odh-manifests-e2e

israel-hdez:
/retest

Tests look unstable... :-/

Jooho:
This PR does not have an integration test script for kserve. This is modelmesh one(https://github.com/opendatahub-io/odh-manifests/blob/kserve-prototype/tests/basictests/modelmesh.sh).

So the e2e test failure has nothing to do with this pr.  I think we can add the test part with another PR after we merge this because we need these manifests as soon as possible.

israel-hdez:
/retest

One more time. If it doesn't pass, I'll find who can help with it.

israel-hdez (approved):

kserve/base/params.env
```diff
@@ -0,0 +1,8 @@
+kserve-controller=quay.io/opendatahub/kserve-controller:v0.10.2
```
Jooho, line 1: @israel-hdez do we need to update this to `latest` or `v0.11-latest` to avoid the anyuid issue?

israel-hdez: Manifests are for 0.10. So, we could use `v0.10-latest`, which should also be patched. Doesn't it work?

Or, do we want to move to 0.11? In upstream 0.11 is still unreleased. So, I'm not sure...

Jooho: I am ok any version that includes the changes regarding anyuid.

LaVLaS:
This appears to be an failure due to the odh operator deploying successfully during the CI test run.  I am going to debug the live system to see what is causing the issue.  Since this PR has no CI tests or modifies the existing functionality of any other components, this should merge regardless of the debug session

LaVLaS:
/approve

Merging this since the ocp411 passed and we have initial confirmation that the ocp412 issues are due to a bug in one of the openshift-ci build images outside of our control.

@israel-hdez @Jooho @ReToCode If there are plans to integrate this a feature into ODH, you'll need to add CI tests when these manifests are updated

PR #846, opened 2023-06-16
Set referencePolicy to source for pulling image directly

harshad16:
Set referencePolicy to source for pulling images directly

## Description

The Source ReferencePolicy allows Imagestream to pull images directly from the external registry.
Without having to rely on the internal ReferencePolicy.
Related-to: https://github.com/opendatahub-io/notebooks/issues/124

## Merge criteria:


harshad16:
Thank you for the approval
Merging this to work on release.

shalberd:
@harshad16 for any non-master branch, referencePolicy of tags should be kept Local, though. Which it is not currently at branch v1.7.

shalberd:
@harshad16 basically, the new image change trigger annotation way of dashboard PR-800 and kubeflow PR 72 will ensure automatically that an internal openshift registry is used, at time of notebook creation from an imagestream tag, when there is an internal openshift registry, and that otherwise, the tag.from.name external location, in sha256 digest format, is used.

PR-72 and PR-800 solve the problem in https://github.com/opendatahub-io/notebooks/issues/124

when there is an internal openshift registry, it is leveraged, when there is not, it is not leveraged.

Now, using tag.referencePolicy: Source is not really necessary anymore, because image policy plugin automatically puts in the external location from i.e. https://github.com/opendatahub-io/odh-manifests/blob/master/notebook-images/base/params.env#L1 in the container image field when no internal openshift docker registry is present, regardless of tag.referencePolicy. That is the beauty of it.

PR #863, opened 2023-07-04
Add monitoring related manifests

zdtsw:
## Description
- imported from odh-deploy's monitoring folder
- when build new operator image from feature-rearchitecture branch of odh-manifests, new operator can get these manifests

## Merge criteria:

### DIFF
--- a/monitoring/alertmanager/alertmanager-svc.yaml
+++ b/monitoring/alertmanager/alertmanager-svc.yaml
@@ -1,0 +1,35 @@
+---
+apiVersion: v1
+kind: Service
+metadata:
+  annotations:
+    service.alpha.openshift.io/serving-cert-secret-name: alertmanager-tls
+  labels:
+    name: alertmanager
+  name: alertmanager
+  namespace: "redhat-ods-monitoring"
+spec:
+  ports:
+    - name: alertmanager
+      port: 443
+      protocol: TCP
+      targetPort: 10443
+  selector:
+    deployment: prometheus
+---
+apiVersion: route.openshift.io/v1
+kind: Route
+metadata:
+  annotations:
+    kubernetes.io/tls-acme: 'true'
+  name: alertmanager
+  namespace: redhat-ods-monitoring
+spec:
+  port:
+    targetPort: alertmanager
+  to:
+    kind: Service
+    name: alertmanager
+  tls:
+    termination: Reencrypt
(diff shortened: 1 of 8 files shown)


etirelli (approved):
LGTM

monitoring/alertmanager/alertmanager-svc.yaml
```diff
@@ -0,0 +1,35 @@
+---
+apiVersion: v1
+kind: Service
+metadata:
+  annotations:
+    service.alpha.openshift.io/serving-cert-secret-name: alertmanager-tls
+  labels:
+    name: alertmanager
+  name: alertmanager
+  namespace: "redhat-ods-monitoring"
```
shalberd, line 10: redhat-ods-monitoring namespace? I don't see such a namespace on any of our clusters.

shalberd:
I stick with my point that introducing such a custom, manually deployed operator for prometheus is something that was ok in OCP 3.11 as a hack, but not anymore in OCO 4.x where other approaches are possible. I know some architect decreed not to use monitoring for user defined projects / user workload monitoring. And I realize that getting to route / haproxy metrics is not immediately obvious, but it's all been discussed before. https://odh-io.slack.com/archives/C03UDTUD0EN/p1686838491101629 I really don't understand what is going on downstream at RHOSDS with regards to monitoring and I fail to understand why one forces this non-recommended way of monitoring stack on upstream. https://docs.openshift.com/container-platform/4.10/monitoring/enabling-monitoring-for-user-defined-projects.html

monitoring/alertmanager/alertmanager-svc.yaml
zdtsw: redhat-* wont exists in any ODH cluster, and the monitoring/*/* will only be applied when detect it as a redhat managed cluster. 
if this makes upstream more confusing, we can apply this change directly in downstream "odh-manifests" repo instead.
the reason to have it here is only to make the sync of odh-manifests from upstream to downstream easier.

shalberd: Yeah, as long as this is an overlay here, it's cool. But I think you all should really get together with your SRE / site reliability engineer, with @VannTen and others to rethink your use of this prometheus here downstream.

zdtsw:
have this one closed, @VaishnaviHire  will make another PR towards downstream only.

PR #905, opened 2023-08-04
[draft] Add queue proxy extension to odh-manifests

Xaenalt:
Add in a field for the queue proxy extension image

Draft, not totally sure what we'll do with it

/hold


VaishnaviHire:
Repo archived

PR #910, opened 2023-08-09
Distributed Workloads v0.1.0 Release

anishasthana:
This updates the Distributed Workloads manifests (CodeFlare and Ray) to the latest as part of the v0.1.0 release. Release notes can be found at https://github.com/opendatahub-io/distributed-workloads/releases/tag/v0.1.0. 

cc @KPostOffice @Maxusmusti @MichaelClifford @tedhtchang @jbusche @LaVLaS


LaVLaS:
Execution of the CI tests is most likely blocked due to an outage in openshift-ci.

jbusche:
Testing looked good - I tried it utilizing this kfdef applied to the 0.8.0 ODH operator:
```
kind: KfDef
metadata:
  name: codeflare-stack
  namespace: opendatahub
spec:
  applications:
  # CodeFlare
  - kustomizeConfig:
      repoRef:
        name: manifests
        path: codeflare-stack
    name: codeflare-stack
  # KubeRay
  - kustomizeConfig:
      repoRef:
        name: manifests
        path: ray/operator
    name: ray-operator
  repos:
    # ODH Core component manifests
  - name: manifests
    #uri: https://github.com/opendatahub-io/odh-manifests/tarball/master
    uri: https://github.com/anishasthana/odh-manifests/tarball/dw_0.1.1
```

jbusche (approved):
LGTM

ray/operator/base/params.env
```diff
@@ -1 +1,2 @@
 namespace=opendatahub
+odh-kuberay-operator-controller-image=quay.io/opendatahub/kuberay-operator:v0.5.0
```
tedhtchang, line 2: Should we use `quay.io/opendatahub/kuberay-operator:v0.6.0`

tedhtchang (commented):
I didn't actually run a e2e test. The change looks good to me. One question, do we need to specify the new kuberay version somewhere?
```
      parameters:
      - name: odh-kuberay-operator-controller-image
         value: docker.io/kuberay/operator:v0.6.0
```
in 
kfdef/codeflare-stack-kfdef.yaml
tests/resources/codeflare-stack/codeflare-stack-kfdef.yaml
kfdef/ray-minimal-kfdef.yaml

ray/operator/base/params.env
anishasthana: I think we should upgrade as part of a PR to DW first, and then copy the manifests in here.

dimakis (approved):
LGTM, tested with a batch job and checked all versions. Everything working as expected.

LaVLaS:
@anishasthana Can you verify the test install script to ensure that the codeflare-operator install is inline with the distributed workloads feature set in this PR?

LaVLaS:
@anishasthana Attached is the pod log from the failed jupyter notebook run during your test
[odh-manifests-PR910-jupyter-nb-kube-3aadmin-0-jupyter-nb-kube-3aadmin.log](https://github.com/opendatahub-io/odh-manifests/files/12341103/odh-manifests-PR910-jupyter-nb-kube-3aadmin-0-jupyter-nb-kube-3aadmin.log)

tests/resources/codeflare-stack/mnist_ray_mini.ipynb
```diff
@@ -50,7 +50,8 @@
    },
    "outputs": [],
    "source": [
-    "cluster.wait_ready()"
+    "cluster.wait_ready()\n",
+    "sleep(30)"
```
anishasthana, line 54: The reason for this is that the ray dashboard takes a few seconds to become accessible, so there is a gap between wait_ready and the job submission being usable. The short term solution is to simply add a small sleep before submitting the job, and the long-term solution (for next codeflare release) will likely involve waiting longer in wait_ready for the dashboard to be accessible, or adding better messaging in this edge case

We will create a follow-on issue for the wait_ready condition. Related, the notebook pods end up restarting almost immediately and then ending up in a crashloop at a different (earlier) point in the notebook. This makes it much harder to debug as the "useful" logs only showed up in the first instance of the crashloop -- every other instance fails due to an unrelated (but not critical) issue.

Thanks @MichaelClifford @Maxusmusti

LaVLaS:
Merging this since it appears to have resolved the Distributed Workloads test.  The modelmesh test is failing and will be fixed in #918 based on an offline conversation by @VedantMahabaleshwarkar

PR #918, opened 2023-08-17
Ensure odh-model-controller compatibility with kserve deployment mode

VedantMahabaleshwarkar:
Odh-model-controller was recently configured to add reconciliation logic for the Kserve Inferencing Stack. This PR makes compatibility related changes to the odh-model-controller manifests to ensure the controller does not break on unexpected conditions. 

## Description

- Manually set `kserve-enabled=false` for the odh-model-controller deployment when used with modelmesh.  
- Add the updated `role` manifests as the new controller image expects these permissions (even though they are not used in this case) 
- Manually add the CRDs for `Telemetries` and `PeerAuthentications`. These are expected to already exist when used with `kserve`. However they are not expected to be installed when used with `modelmesh`. Since the controller logic expects the CRDs to be installed we manually add them to the manifests here. 

## How Has This Been Tested?

- Install Opendatahub operator from the `fast` channel 
- Add the following to the `DSCInitialization` CR as follows `manifestsUri: 'https://github.com/VedantMahabaleshwarkar/odh-manifests/tarball/disable-kserve-for-mm'`
- Create a `DataScienceCluster` CR with only `modelmeshserving.enabled=true` 
- Test the creation of a onnx isvc 

## Merge criteria:


heyselbi (approved):

taneem-ibrahim:
The modelmesh tests can be ignored based on @VedantMahabaleshwarkar retesting .

